Issue #62, opened 2016-02-09
Refactor grounding information representation/access


hickst:
The implementation details of the Grounding trait are spreading throughout the Reach software, requiring every access to understand the structure of the representation it is reaching into (violates the "Law of Demeter"). This makes it very difficult to change the representation, which is currently necessary as more information is required to be returned and represented.

Issue #59, opened 2016-02-04
Running Reach several times produces different results


hickst:
Some documents producing more or fewer results over consecutive runs. (ex: PMC3441633).

hickst:
Ran ReachCLI on PMC3441633.nxml twice. Second run produced additional event mention, missing from first run:
```> ================================================================================
> CorefEventMention: S3/723/768: Positive_activation
> text: Ras activation and engagement of specific Ras
> labels: List(Positive_activation, ActivationEvent, ComplexEvent, Event, PossibleController)
> foundBy: Positive_activation_syntax_3_noun
> trigger:
>   TextBoundMention: S3/727/737: Positive_activation
>   text: activation
>   labels: List(Positive_activation, ActivationEvent, ComplexEvent, Event, PossibleController)
>   foundBy: Positive_activation_syntax_3_noun
> controlled (1):
>   CorefTextBoundMention: S3/765/768: Family
>   text: Ras
>   labels: List(Family, BioChemicalEntity, Entity, PossibleController)
>   foundBy: ner-family-entities
>   xref: Xref(interpro,IPR020849)
> controller (1):
>   CorefTextBoundMention: S3/723/726: Family
>   text: Ras
>   labels: List(Family, BioChemicalEntity, Entity, PossibleController)
>   foundBy: ner-family-entities
>   xref: Xref(interpro,IPR001806)
> context:
>   Cellular_component: List(go:GO:0005623)```

Copied 4th sentence (S3 above) and pasted it into ReachShell 3 times: first and second time the event was produced. The third time NO event was produced. Sentence is:

```Our findings suggest a novel model for compartmentalized signaling in which PI3KC2&#x3B2; results in the redistribution of nucleotide-free Ras to intracellular vesicles leading to localized Ras activation and engagement of specific Ras effectors resulting compartmentalized activation of these targets.```

myedibleenso:
@hickst is it only the event that's missing?  Are the entities being found all three times?

hickst:
For ReachCLI runs, the answers are (diff file attached):
[run1-2_diff.txt](https://github.com/clulab/reach/files/118145/run1-2_diff.txt)

1) for the two different 2 runs, yes (run 1 results were the same as run 3 results)
2) yes - no entity changes
3) yes - no entity label changes

For the test at the ReachShell, the answers are:
1) yes, (but it's a single sentence)
2) yes - no entity changes
3) yes - no entity label changes.

myedibleenso:
[...]

PR #64, opened 2016-02-12
Ground4

hickst:
Resolves issue #62.


hickst:
Forgot to mention, this code also fixes issue #59.

Issue #65, opened 2016-02-13
Hardwired Grounding namespace values used in protein/family tests in ReachSystem


hickst:
ReachSystem (*outside of the grounding package*) contains several instances of *hardwired* IF tests using *grounding namespace values*. This is a mysterious error waiting to happen if/when we change the grounding KBs we use (or their namespaces).

PR #66, opened 2016-02-13
Issue 65: Implement better "is family/protein grounded" methods, add …

hickst:
Fixes issue #65. This request builds upon work in PullRequest #64. #64 **must** be applied **before** this pull request.


hickst:
Now that #64 is merged, I'll resubmit this request against the master branch.

PR #68, opened 2016-02-16
Catch binding in 'X that is bound by Y'

danebell:
I added a binding rule to match examples from the assembly requirement PPT such as:
> Src tyrosyl phosphorylates Ras that is GTP bound

with the variation 
> Ras that is GTP bound is phosphorylated by Src tyrosyl

which has slightly different syntax because of parser issues.

I also added unit tests for each of these sentences. Other tests still pass.

### DIFF
--- a/src/main/resources/edu/arizona/sista/reach/biogrammar/events/bind_events.yml
+++ b/src/main/resources/edu/arizona/sista/reach/biogrammar/events/bind_events.yml
@@ -317,6 +317,16 @@
   pattern: |
     [lemma=/^(when|after|once|if)/] (?<trigger> [lemma="bind"]) [lemma="to"]? [tag=/^(DT|JJ)/]* @theme1:BioChemicalEntity [tag=/^NN/]? ","? [tag=/^(DT|JJ)/]* @theme2:BioChemicalEntity
 
+- name: binding_thatisbound
+  label: Binding
+  action: mkBinding
+  priority: 4
+  example: "Src tyrosyl phosphorylates Ras that is GTP bound"
+  pattern: |
+    trigger = [word=bound & tag=/^VB(D|N)$/]
+    theme1:BioChemicalEntity+ = (<rcmod | <vmod)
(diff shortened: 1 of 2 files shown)


myedibleenso:
This looks good for a quick fix, @danebell, but I'm wondering if this is something we need to address more comprehensively.  This sentence could've been "Src tyrosyl phosphorylates Ras that is ubiquitinated" or "Ras that was previously ubiquitinated".  

Here is what is currently captured for "Src tyrosyl phosphorylates Ras that is ubiquitinated":

```
mention text: Src tyrosyl phosphorylates Ras
List(Positive_regulation, Regulation, ComplexEvent, Event, PossibleController)
	Rule => Phosphorylation_syntax_1_verb
	Type => CorefRelationMention
	controlled (List(Phosphorylation, SimpleEvent, Event, PossibleController)) => phosphorylates Ras
	controller (List(Gene_or_gene_product, MacroMolecule, BioChemicalEntity, Entity, PossibleController)) => Src

mention text: Ras that is ubiquitinated
List(Ubiquitination, SimpleEvent, Event, PossibleController)
	Rule => Ubiquitination_syntax_6_verb
	Type => CorefEventMention
	trigger => ubiquitinated
	theme (List(Family, BioChemicalEntity, Entity, PossibleController)) => Ras
```

I'm wondering if that relative clause should actually be captured as a PTM on the entity, so that we see it in the `Phosphorylation`.  What do you think, @MihaiSurdeanu?

danebell:
I was skeptical that there were examples like what you site, Gus, but I've found several, e.g. `KEAP1 interacts with endogenous WTX that is phosphorylated at serine 286` I can write a similar PTM rule.

myedibleenso:
Yeah, I think we'll have to capture this information somewhere.  I think you know the drill with the PTM rules:  A PTM rule should only match when a cause isn't present.  Otherwise we'll run into problems (ex. a Phosphorylation of a phosphorylated entity).  Still waiting for @MihaiSurdeanu to weigh-in...  

After that, I will close this pr and open an issue for extending PTMs.

PR #77, opened 2016-02-23
Refined GGP and Family NER rules

### LINKED ISSUE #74: Change the Odin BIO rules: if it ends with "family", etc. label as Family instead of Protein.
Since grounding now dispatches on head label, the bio rules should recognize, and set the appropriate label, for the special cases:

XXX protein family
XXX family

Once rules are enhanced, grounding code will be updated to remove the several special cases.

myedibleenso:
This PR resolves #74.  The improvement uses a lookahead assertion on the token that follows the standard ner rule of the form `[entity=B-somelabel] [entity=I=somelabel]*`). 

### Changes
- A token sequence labelled as GGP by the crf will produce a mention with the label `Family` if the lemma form of the token that follows is "family".  

- A similar check prevents a `GGP` mention being created if a token sequence labelled as GGP by the crf is followed by a token which has the lemma "family".

### Problems
While this doesn't seem to break anything, this change deserves a test or two.  I'd appreciate any examples of strings that are consistently being labeled GGP or Family.

### DIFF
--- a/src/main/resources/edu/arizona/sista/reach/biogrammar/entities/entities.yml
+++ b/src/main/resources/edu/arizona/sista/reach/biogrammar/entities/entities.yml
@@ -141,7 +141,10 @@
   priority: 3
   type: token
   pattern: |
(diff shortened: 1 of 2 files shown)


MihaiSurdeanu:
@hickst: any examples for entities currently labeled as GGP, when they contain "family"?

myedibleenso:
@hickst, did you ever come across any examples that I could use for tests?

myedibleenso:
I've added a test for this change and things appear to be working.

Issue #72, opened 2016-02-19
Enumerate the KB contents in memory


enoriega:
Context engines build a global vocabulary of context types by reading KB files from bio resources and concatenating the lists of IDs in them (the last item on each line of a tsv file).

To recover the context index of a mention in the context vocabulary, we rely on grounding. With the current refactor of grounding, the grounding id is only accessible with via the `nsId` field. This field concatenates the name sapace.

The KB files don't include the namespace in their text, so I use the ad hoc approach of splitting the nsId with `:` and take the last field. I.E. `taxonomy:9606` becomes `9606` which matches the id in `Species.tv.gz`

The problem with this is that _it's a hack_ and the format of the ids in the files is inconsistent, for example, in the KB `GO-subcellular-locations.tsv.gz` the ids have the form 	`GO:0005856` and this breaks my ad hoc approach.

I believe that the vocabularies should contain the namespace to avoid id collisions of different files. An ideal solution would be to have a method that roughly does something like this:

`def getKBEntries(kbName:String) = foreach line in contents(kbName) map { (nsId, text) }`

Where `nsId` is the namespace + id as in the grounding trait and `text` is one of the strings that represent that id.

hickst:
The current refactoring of grounding provides access to **separate** namespace, id, text, key, and species fields (all Strings), AND gives you a convenient nsId method for combining the namespace and the ID. Please read the source code for grounding/KBResolution.scala to see what fields are available.

enoriega:
Right, but this is for the resolution of a given mention. I need to enumerate the contents of a KB file. If I just use the id field in the KBResolution instance of a mention, I'm still in risk of colliding with entries of different files with the same id, as there is no way to get the prefix from the KB tsv.

hickst:
So instead of starting with a total enumeration of all Context KB entries, why don't you build up a data structure containing just the context groundings actually found: adding to the data structure after each file is processed? I'm pretty sure that such a data structure will contain about 10% (or less) of all the possible groundings. Since each grounding resolution includes all the fields I enumerated above, you can separate groundings from different namespaces that may have the same ID.

enoriega:
No, because this is used to build an index of context entities that should be consistent across papers and across run. Building an index on the fly will contain only entries for the mentions from that paper and won't be consistent.

This was they way it was before and we can't compare papers or do learning without a global dictionary

hickst:
ENor says: current solution is missing the Cellular_Component related KBs:
[...]

PR #78, opened 2016-02-23
Accumulated grounding work.

hickst:
Add enumeration of KB entries (issue #72)
Fixes bioprocess grounding bug (issue #75)
Moves finding of grounding candidates to after modifications.
Enhancements to textual output, and grounding trait.
Numerous cleanups.
Adds test suite for Grounding trait.


MihaiSurdeanu:
Is this where the context tests fail?

hickst:
Maddening thing is that sometimes it fails, sometimes it works. But, since it has failed and sometimes continues to fail on the current master branch, I figure I would go ahead and issue the pull request.

MihaiSurdeanu:
In ReachSystem, I see a new method: mutationsToMentions. Does this conflict with @danebell's work in coref? @hickst and @danebell: can you please double check?
Otherwise, I think it's ready to merge.

hickst:
mutationsToMentions is not new code: I just pulled it out of a function above it as it was a self contained piece of logic.

PR #80, opened 2016-02-23
Alias

danebell:
This is a first attempt at having entity mentions share grounding when only one has good grounding and the text indicates that the mentions are synonymous. This sharing continues across sentences. For example, "Akt (also called Akt334, AktTR, or Akt4H) is phosphorylated. AktTR is also ubiquitinated." now produces five entity mentions, all with identical grounding from `Akt`.

This is accomplished by means of rule-based collection of RelationMentions with an `Alias` label that are used by Coref to share grounding and then filtered out.

Limitations: 
* This doesn't work yet for the small number of coordinated examples, e.g. "MLL1/MLL2 (also known as KMT2A and KMT2B, respectively)"
* This only works for mentions tagged Family and MacroMolecule. An example such as "Ras S135M (RasM)" won't do this sharing because the mutation is getting in the way.
* The two mentions must have the same label. This is problematic for those times when they actually are the same, but for NER reasons have been given different labels, but it seems better to be conservative at first.

### DIFF
--- a/src/main/resources/edu/arizona/sista/reach/biogrammar/coref/alias.yml
+++ b/src/main/resources/edu/arizona/sista/reach/biogrammar/coref/alias.yml
@@ -7,7 +7,7 @@
     (@alias:Family | @alias:MacroMolecule)
     [tag="-LRB-"]
     (@alias:Family | @alias:MacroMolecule)
-    ([word=/^,|or$/]+ (@alias:Family | @alias:MacroMolecule))*?
+    ([word=/^,|\/|or$/]+ (@alias:Family | @alias:MacroMolecule))*?
     [tag="-RRB-"]
 
 - name: alias_paren2_referredto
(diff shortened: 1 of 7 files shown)


MihaiSurdeanu:
Dane, can you please use a single static variable for "uaz", instead of copying the string "uaz" throughout the code? Just in case we end up using a different prefix for underdefined grounding.
Otherwise, this is ready to merge.

hickst:
If "uaz" is meant to be a default namespace string, you can use the class constant DefaultNamespace in
grounding/ReachKBConstants (whose value happens to be "uaz").

MihaiSurdeanu:
@danebell: can you please follow Tom's suggestion?

danebell:
Okay, the only use of the actual string "uaz" is in ReachKBConstants now.

Issue #81, opened 2016-02-24
Grounding should handle protein-domain strings


hickst:

hickst:
Use the domain string to identify and ground the protein.

hickst:
In the future, we could make the pattern match smarter but I'm not even sure how frequently the basic PPP-DDD pattern occurs so, for now, perhaps we can close this issue.

MihaiSurdeanu:
I agree.

PR #85, opened 2016-02-26
Use the protein-domain string to identify and ground the protein.

hickst:
Implements feature request of issue #81.


hickst:
No, in fact, this simple pattern only captures  PPP-DDD. It would be better to enhance the pattern to capture (PPP)+-DDD since many proteins have more than one hyphenated section: (e.g. PI3K-C2-alpha).

PR #91, opened 2016-02-29
Issue90: grounding pXXX => XXX

hickst:
Implements PTM-related prefix stripping as alternate lookup method for proteins. 
This simple approach should be very effective but, because of current grounding design and data limitations, has the remote potential to create some false positive groundings (i.e. stripping a PTM-prefix from a string could potentially result in an incorrect protein identification). This should be very unlikely, especially as this algorithm is run last in the series of key lookups.


MihaiSurdeanu:
I wonder if we need to add other patterns in addition to p|u? Can you please check with RGX?
We also need to discuss how we can recover the PTM later in Reach.
I will merge this for now.

PR #93, opened 2016-03-01
Fix typo in pom.xml dependency stanza in README

jmuhlich:
The groupId must be `org.clulab`, not `clulab.org`.


MihaiSurdeanu:
Thanks!

PR #94, opened 2016-03-01
Update paxtools-core dependency to 4.3.1

jmuhlich:
BioPAX is now officially releasing to Maven Central as of PaxTools 4.3.1, so
let's make use of that. (See https://groups.google.com/d/msg/biopax-discuss/zwtwDG23T1E/Vu1OK7iXBQAJ for the release announcement mentioning Maven Central)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -75,7 +75,7 @@ libraryDependencies ++= Seq(
   "org.clulab" %% "processors" % "5.8.1" classifier "models",
   "com.typesafe" % "config" % "1.2.1",
   "commons-io" % "commons-io" % "2.4",
-  "org.biopax.paxtools" % "paxtools-core" % "4.3.0",
+  "org.biopax.paxtools" % "paxtools-core" % "4.3.1",
   "jline" % "jline" % "2.12.1",
   "org.apache.lucene" % "lucene-core" % "5.3.1",
   "org.apache.lucene" % "lucene-analyzers-common" % "5.3.1",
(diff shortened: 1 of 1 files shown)


myedibleenso:
Thanks, @jmuhlich.  I think that means we should get rid of this, too: 

https://github.com/jmuhlich/reach/blob/5d0d3486901dfe92d6065bbed7f070644c2454f4/build.sbt#L66-L69

jmuhlich:
They mention that snapshots will be available at https://oss.sonatype.org/content/groups/public/ . Does that need to be included in `resolvers` explicitly?

myedibleenso:
We're not really making use of snapshots for this dependency, but I suppose it doesn't hurt.  

Based on [the sbt reference](http://www.scala-sbt.org/0.13/docs/Resolvers.html), I think the block I linked to could be replaced with this line: 

````scala
resolvers += Resolver.sonatypeRepo("public")
```

jmuhlich:
`Resolver.sonatypeRepo("public")` becomes `https://oss.sonatype.org/content/repositories/snapshots/` which is different from the URL given in the google groups post, `https://oss.sonatype.org/content/groups/public/`. Both have `org/biopax/paxtools/paxtools-core` but the contents are different. I am not a Java guy so I don't know what to make of this.

myedibleenso:
Oh, thanks!  I just tried several different resolvers, but none of them were able to find a `4.3.1-SNAPSHOT`.

We really don't need snapshot versions for this dependency, so let's just keep it simple and remove those resolvers.

jmuhlich:
OK, pushed another commit removing `resolvers`.

PR #109, opened 2016-03-09
Control the number of threads used in parallelization

myedibleenso:
I've added a variable (`threadLimit`) under `application.conf` to control the size of the thread pool used in parallelization.  So far I've only integrated `threadLimit` into the `ReachCLI` and `PaperReader` runnables.


myedibleenso:
In the future, I think we should wait until travis finishes before merging.

PR #134, opened 2016-03-28
deleted wrong pattern fragment

marcovzla:
Deleted a fragment of a pattern that isn't doing anything other than convoluting it.
The word "by" shouldn't be reachable using collapsed dependencies.
If it is reachable, it is probably not part of the controlled argument.
I'm not sure what was the intention of whoever wrote this, but I suspect it is not doing that.


myedibleenso:
Good catch.  The lab should devote a day or two over the summer to cleaning up the grammars.  There are a few spurious things floating around.  We don't want to point Odin neophytes to our grammars only to confuse them.

PR #158, opened 2016-04-14
Change Reach to use PFAM protein family KB

hickst:
*This release depends on Bioresources version 1.1.4 so tests will not pass until that is available.*

Added PFAM protein family KB with InterPro retained as secondary family KB. Modified lookups, updated tests, and fixed one site rule in entities.yml.


hickst:
Closing pull request to restart Travis build.

Issue #113, opened 2016-03-11
Extract dephosphorylation


bgyori:
As far as I see, there is currently no mechanism to extract dephosphorylation events. For instance, parsing the sentence "The phosphatase DUSP6 dephosphorylates MAPK1 at threonine 185 and tyrosine 187."  returns no events.

MihaiSurdeanu:
Ben,
This is trivial to add. We just didn't know this was relevant.
Mihai

bgyori:
Yes, I think it is relevant - it is as important as phosphorylation but less well understood. Still, it is commonly discussed in the literature and in RAS biology in particular.

hickst:
Hi Ben,

While we're adding dephosphorylation, which other PTMs events might be important to add? Any or all of the following? Are there ones I've missed which should be included? Thanks for your help!

Deacetylation
Defarnesylation
Deglycosylation
Dehydrolysis
Dehydroxylation
Demethylation
Deribosylation
Desumoylation
Deubiquitination

bgyori:
I think that list is pretty comprehensive. 99% of the time our group will be interested in Dephosphorylation only but - as long as it's not a lot of extra work - the others listed can definitely be included. I can't think of any further ones right now that are important.

hickst:
Thanks Ben! We'll get to work on adding these.

myedibleenso:
## Proposal
Here are the components I know that we'll need to change:

- [ ] the [`taxonomy`](https://github.com/clulab/reach/blob/f0db436a0efba480b1b6440a2ccfac4a745c478e/src/main/resources/edu/arizona/sista/reach/biogrammar/taxonomy.yml#L26-L39) should be modified to include De* events under `SimpleEvents`.  Ideally, these should be nested under some category under `SimpleEvent`.  Doing so will make things easier in assembly...
  - ex.
   ```
    - SimpleEvent:
      - RemovalEvent:
        -Dephosphorylation
        ...
      - Phosphorylation
     ....
  ```

[...]

PR #159, opened 2016-04-18
Issue113: Add "reverse" PTM events

hickst:
Answers issue #113: Extract dephohphorylation (and other de-events).

### DIFF
--- a/src/main/resources/edu/arizona/sista/reach/biogrammar/events_master.yml
+++ b/src/main/resources/edu/arizona/sista/reach/biogrammar/events_master.yml
@@ -21,12 +21,13 @@ rules:
   # Phosphorylations
   - import: edu/arizona/sista/reach/biogrammar/events/simple-event_template.yml
     vars:
-     eventName: "Phosphorylation"
-     actionFlow: "mkBioMention"
-     label: "Phosphorylation"
-     verbalTriggerLemma: "phosphorylate"
-     nominalTriggerLemma: "phosphorylation"
-     priority: "5"
+      eventName: "Phosphorylation"
+      actionFlow: "mkBioMention"
+      label: "Phosphorylation"
+      verbalTriggerLemma: "phosphorylate"
+      nominalTriggerLemma: "phosphorylation"
+      deventPrefix: "!word=/(?i)^de/"
+      priority: "5"
 
   # Ubiquitination
   - import: edu/arizona/sista/reach/biogrammar/events/simple-event_template.yml
@@ -36,6 +37,7 @@ rules:
       label: "Ubiquitination"
       verbalTriggerLemma: "ubiquitinate"
       nominalTriggerLemma: "ubiquitination"
+      deventPrefix: "!word=/(?i)^de/"
       priority: "5"
 
   # Hydroxylation
@@ -46,6 +48,7 @@ rules:
       label: "Hydroxylation"
       verbalTriggerLemma: "hydroxylate"
       nominalTriggerLemma: "hydroxylation"
+      deventPrefix: "!word=/(?i)^de/"
       priority: "5"
 
   # Sumoylation
@@ -56,6 +59,7 @@ rules:
       label: "Sumoylation"
       verbalTriggerLemma: "sumoylate"
       nominalTriggerLemma: "sumoylation"
+      deventPrefix: "!word=/(?i)^de/"
       priority: "5"
 
   # Glycosylation
@@ -66,6 +70,7 @@ rules:
       label: "Glycosylation"
       verbalTriggerLemma: "glycosylate"
       nominalTriggerLemma: "glycosylation"
+      deventPrefix: "!word=/(?i)^de/"
       priority: "5"
 
   # Acetylation
@@ -76,6 +81,7 @@ rules:
       label: "Acetylation"
       verbalTriggerLemma: "acetylate"
       nominalTriggerLemma: "acetylation"
+      deventPrefix: "!word=/(?i)^de/"
       priority: "5"
 
   # Farnesylation
@@ -86,6 +92,7 @@ rules:
       label: "Farnesylation"
       verbalTriggerLemma: "farnesylate"
       nominalTriggerLemma: "farnesylation"
+      deventPrefix: "!word=/(?i)^de/"
       priority: "5"
 
   # Ribosylation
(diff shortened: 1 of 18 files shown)


MihaiSurdeanu:
This looks good. 
But what are all those de* trigger rules in generic_events.yml? Are we actually using them?

src/main/resources/edu/arizona/sista/reach/biogrammar/events_master.yml
```diff
@@ -106,8 +118,113 @@ rules:
...
+      triggerPrefix: "word=/(?i)^de/"
+      priority: "5"
+
+  # Desumoylation
+  - import: edu/arizona/sista/reach/biogrammar/events/simple-event_template.yml
+    vars:
+      eventName: "Desumoylation"
+      actionFlow: "mkBioMention"
+      label: "Desumoylation"
+      verbalTriggerLemma: "de-?sumoylate"
```
myedibleenso: If you're using the `triggerPrefix`, why do you need to include the "de-?" pattern in the `verbalTriggerLemma` and `nominalTriggerLemma` patterns?

danebell:
@MihaiSurdeanu The trigger rules in generic_events.yml make sense to me. This is for cases like the following:

> A desumoylates B. This desumoylation is promoted by C.

The second sentence has a generic event whose antecedent is the desumoylation in the first sentence. Probably pretty rare, but logically possible.

src/main/resources/edu/arizona/sista/reach/biogrammar/events_master.yml
hickst: Note that the trigger prefix, as per your suggestion, is the matching expression string "word=/(?i)^de/" and not the actual prefix string itself. Using the actual prefix string alone as a parameter is not possible since the matching *rule* needs to be modified, not the pattern it is matching.

src/main/resources/edu/arizona/sista/reach/biogrammar/events_master.yml
```diff
@@ -106,8 +118,113 @@ rules:
...
+  # Deglycosylation
+  - import: edu/arizona/sista/reach/biogrammar/events/simple-event_template.yml
+    vars:
+      eventName: "Deglycosylation"
+      actionFlow: "mkBioMention"
+      label: "Deglycosylation"
+      verbalTriggerLemma: "de-?glycosylate"
+      nominalTriggerLemma: "de-?glycosylation"
+      triggerPrefix: "word=/(?i)^de/"
+      priority: "5"
```
myedibleenso: No, you don't want to encode the prefix in two places.  The `verbalTriggerLemma` and `nominalTriggerLemma` should be identical to the corresponding additive event.  For example, "deglycosylation" and "glycosylation" should have the same `verbalTriggerLemma` and `nominalTriggerLemma`; the distinction can be made in the `triggerPrefix`.

hickst: After chatting offline, I think I understand your proposal but I still don't agree with it for the following reasons:
1) I don't see why they should be identical to the corresponding additive event -- they are used in two different rule sets and the change doesn't save anything at all,
2) we are not encoding the prefix in two places: the lemma trigger and the triggerPrefix are not the same pattern; they are not even both patterns: one is a pattern but the other is a "rule fragment" containing a (slightly) different pattern.
3) the pattern and the fragment might be combined but this would be less efficient because:
   a) the template rules are based on a conjunction of tests and
   b) the lemma pattern is always tested first and
   c) the current lemma pattern (de-?) is more efficient because it narrows/restricts the set of lemmas matched by the first match (in the template rules), eliminating the need for the second match (in the template rules) in any case where the word contains phosphorylation (e.g. oxidative-phosphorylation)
4) The lemma prefixes also document what that rule is looking for....whereas using just "phosphorylation" is confusing, since the rule is specifically *not* looking for phosphorylation.

myedibleenso:
I see you already have merged, but I wanted to respond to your last comment.

1.  You're right that there is no law that says we must make these two forms (additive and removal variants) as similar as possible, it's just that *they only differ in the trigger's prefix* and these rules have a variable added to capture exactly that difference (`triggerPrefix`).

2.  It sounds like there is still some confusion on this point...

3c. Yes, the *initial* check you decided to use is more efficient, but then you end up making basically the same check with the next constraint on the `triggerPrefix`.  If we're talking about efficiency, we should really have a set of trigger rules which runs once prior to the event rules.  Then the event templates could be further simplified.

>since 1/2 of the template-instantiated rules are specifically not looking for phosphorylation.
I'm not sure I understand this, but I think you mean that the corresponding "removal" event to a phosphorylation is not a "phosphorylation", but a "de-phosphorylation".

Yes, what you merged here works:
```yaml
- import: edu/arizona/sista/reach/biogrammar/events/simple-event_template.yml
  vars:
    eventName: "Deglycosylation"
    actionFlow: "mkBioMention"
    label: "Deglycosylation"
    verbalTriggerLemma: "de-?glycosylate"
    nominalTriggerLemma: "de-?glycosylation"
    triggerPrefix: "word=/(?i)^de/"
    priority: "5"
```

I just found this simpler:
```yaml
- import: edu/arizona/sista/reach/biogrammar/events/simple-event_template.yml
  vars:
    eventName: "Deglycosylation"
    actionFlow: "mkBioMention"
    label: "Deglycosylation"
    verbalTriggerLemma: "glycosylate"
    nominalTriggerLemma: "glycosylation"
    # the trigger's prefix tells us this is a "removal" event
    triggerPrefix: "word=/(?i)^de-?/"
    priority: "5"
```

Different strokes for different folks?

hickst:
1) But the additive and removal rules *don't* differ only in their lemma strings -- they assign a different label, which means we still need both rules. Therefore, in the current solution:
2) the different lemma strings act as documentation, at the point of call, of what the rule is matching. It seems downright misleading for the lemma triggers to say "glycosylation" when the rule is built to match, and will only ultimately match, "de-?gylcosylation"
AND
3) it is more efficient because I *explicitly avoid making the second check* for a larger set of items than your variation. Your variation must make checks 1 AND 2 for ALL items containing "glycosylation", whereas mine makes check 2 ONLY for items which contain "de-?glycosylation", a more restricted set.

(UNLESS you are telling me that a conjuction of patterns (eg., /p1/ & /p2/ & /p3/) is not implemented with a "short-circuit logical AND" in Odin?) in which case neither of our variants is more efficient.

PR #164, opened 2016-04-20
Issue149: Add tests for species grounding in 2nd grounding layer

hickst:


MihaiSurdeanu:
"human groundings" I wonder what non-biologists reading our code think...

PR #165, opened 2016-04-21
Issue135: track if regulation is direct or not

hickst:
Added boolean flag 'isDirect' to BioEventMentions. Added output of flag to text, FRIES, and indexcard output formats. Added some tests of flag.

### DIFF
--- a/src/main/scala/edu/arizona/sista/reach/extern/export/MentionManager.scala
+++ b/src/main/scala/edu/arizona/sista/reach/extern/export/MentionManager.scala
@@ -15,7 +15,7 @@ import edu.arizona.sista.reach.mentions._
 /**
   * Defines methods used to manipulate, cache, and output Mentions.
   *   Written by Tom Hicks. 4/3/2015.
-  *   Last Modified: Update to use Reach constants object.
+  *   Last Modified: Output new isDirect event field.
   */
 class MentionManager {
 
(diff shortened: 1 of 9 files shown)


src/main/scala/edu/arizona/sista/reach/extern/export/MentionManager.scala
```diff
@@ -140,6 +140,8 @@ class MentionManager {
         }
 
       case evm: EventMention =>
+        if (evm.isInstanceOf[BioEventMention])
+          mStrings += s"${indent}is-direct: ${evm.asInstanceOf[BioEventMention].isDirect}"
```
myedibleenso: Instead of this, maybe it would be cleaner to pattern match on `mention.toBioMention` and change each of these cases to their bio counterpart (ex. `case evm: EventMention` => `case evm: BioEventMention`).  If you match against the `Bio` form, you won't need this `isInstance` & `asInstance` stuff.

myedibleenso:
@hickst, I forgot to mention this earlier, but can you update the changes section of the readme with this enhancement?  Also, we forgot to update it when the Addition/Removal events stuff was merged.  Both of these are worth mentioning there, I think, especially since they were requested features.

hickst:
Updated CHANGES file with comment about this change. The A/R stuff was already in the CHANGES file, albeit not using those terms.

Issue #168, opened 2016-04-27
Update FRIES output for AssemblyEval


hickst:
- [x] 1. Evaluate new IC Schema -- feedback
- [x] 2. Identify changes to FRIES:
    * a) renames, reformats, restructuring?
    * b) new structures required?
    * c) missing information?
- [ ] 3. Modify FRIES output

hickst:
Notes:
1. Context types still not arrays in schema (context is, types aren't)
2. :exclamation: Genes vs. Proteins vs. Families
3. :exclamation: Binds vs. Complexes_with
4. :interrobang: Do we need to populate new protein family structure?
5. :question: Entity text combined "across the text" - what is scope of this?

hickst:
Update FRIES output to copy location features from context to Location participant feature of event arguments.

hickst:
Many things have been done on this task but this issue is too broad and open-ended so closing this.

PR #171, opened 2016-05-04
Add new context frames and pointers to FRIES output event stream

hickst:
Implements context handling from Hans' v11 schema. One part of issue #168.


MihaiSurdeanu:
Merged. But we should start adding tests for these exporters...

PR #176, opened 2016-05-09
updates to assembly

myedibleenso:
This PR includes the following changes to assembly:
- new rule-based sieves for assembly
- an assembly shell for viewing de-duplicated events and their causal precedence
- utilities to train and evaluate a causal precedence classifier
- utilities to read and write an assembly relation corpus
- minor improvements to the `AssemblyManager`
- refactoring of assembly components


myedibleenso:
It also includes some tests for the rule-based sieves

PR #205, opened 2016-05-23
dependency updates

antonkulaga:
I added dependencyUpdates plugin and updated several dependencies and also version of sbt ( sbt 3.11 is less buggy, resolves faster because of caching and allows to add useful compiler plugins in the future), I also added scalastyle that allows to highlight code style issues in IDEs

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ version := "1.2.3"
 
 organization := "org.clulab"
 
-scalaVersion := "2.11.6"
+scalaVersion := "2.11.8"
 
 scalacOptions ++= Seq("-feature", "-unchecked", "-deprecation")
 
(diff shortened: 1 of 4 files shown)


MihaiSurdeanu:
@marcovzla, @hickst: can you please review this pull request?
Not sure why we need the scalastyle-config file?

hickst:
CLULAB: before accepting:
1) We would need to test this branch thoroughly to insure compatibility across our other projects, especially Bioresources and Processors.
2) We should update Processors dependencies to match these updates, if compatible
3) Based on our principle to minimize dependencies for users, I am inclined to exclude the plugins: users can add these for themselves.

antonkulaga:
Hi @hickst 

As it is my first PR here I want to clarify some details:

> 1) We would need to test this branch thoroughly to insure compatibility across our other projects, especially Bioresources and Processors.

I tested only against tests inside Reach project, so I cannot guarantee that there will be no issues with tests in projects that depend on Reach. Are there any tests outside of Reach project that I should run each time I make a PR?

>3) Based on our principle to minimize dependencies for users, I am inclined to exclude the plugin

As far as I understand sbt plugins do not leak to published maven artifacts.

antonkulaga:
>Not sure why we need the scalastyle-config file?

Maybe you do not. As I only started learning the codebase I've made the most simple PR possible - build configuration update. scalastyle is a common practise to keep the same coding style inside the team: if you make things like extra spaces, variable duplications and so on, then you get a warning

myedibleenso:
@antonkulaga, thanks for the contribution.  Would you please remove the `scalastyle` stuff?  We're still discussing how we want to handle code style.

antonkulaga:
@myedibleenso

Ok, I can remove it and rebase on top of the current master, is there anything else I should do?

myedibleenso:
No, that should do it.  Thanks!

antonkulaga:
I am on the trip now right now, when I will return I will try to rebase and fix the PR

myedibleenso:
@antonkulaga, I'm closing this for now, but feel free to open a new one when/if you have time to contribute the discussed changes (dependency updates, right?).

PR #209, opened 2016-05-24
Issue190: Replace homemade KBs

hickst:
This branch replaces the local cell-type KB with Cell Ontology, local cell line KB with Cellosaurus, and local organ KB with Uberon. NER files were then updated. 

This branch increments version number to 1.1.7.


hickst:
Sorry, wrong project -- ignore this pull request: I will re-issue when the dependent projects have been updated.

PR #229, opened 2016-06-09
'is dependent on' precedence rule

### LINKED ISSUE #186: Add "is dependent on" (other variants?) to precedence rules
Together these data demonstrate that E2-induced SRC-3 phosphorylation is dependent on a direct interaction between SRC-3 and ERα and can occur outside of the nucleus.

danebell:
This commit closes #186 by handling the construction "EventA is dependent on the EventB" with both dependency and surface rules. I took the liberty of reorganizing our assembly test files as well.


myedibleenso:
Excellent (insert Burns gif).  Thanks for cleaning up everything as well!

PR #231, opened 2016-06-09
Issue 230: X inhibitor -> Simple_chemical

### LINKED ISSUE #230: "Protein|Family inhibitor|inhibitors" -> Simple_chemical
@MihaiSurdeanu :
>Add drug NE rules in Reach for "Raf Inhibitors". That is: "Protein|Family inhibitor|inhibitors" should be labeled as Drug.

"Raf inhibitors" should match as a Simple_chemical, and the "Raf" should not match as a Family (or anything else).

There are many protein names that include the substring `inhibitor` e.g. `Venom chymotrypsin inhibitor`, so this should only apply to cases where `inhibitor` is not part of an existing entity.

danebell:
This closes #230 by ensuring that (potentially multi-word) protein or protein family names followed by "inhibitor" are marked as Simple_chemicals, and that the creation of a mention for the protein or family entity itself is blocked. Tests are provided.

### DIFF
--- a/src/main/resources/edu/arizona/sista/reach/biogrammar/entities/entities.yml
+++ b/src/main/resources/edu/arizona/sista/reach/biogrammar/entities/entities.yml
@@ -150,14 +150,16 @@
   priority: 3
   type: token
   pattern: |
-    [entity='B-Gene_or_gene_product' & !word=/(?i)^([ACDEFGHIKLMNQRSTVWY]?\d+_|Delta)?[ACDEFGHIKLMNQRSTVWY]?\d+(del|ins|dup|fs)?[ACDEFGHIKLMNPQRSTVWY]*$/]+ [entity='I-Gene_or_gene_product']* (?! [lemma="substrate"])
-    [entity='I-Gene_or_gene_product']*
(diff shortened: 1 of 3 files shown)


myedibleenso:
I'm restarting the build.  I can't understand why it failed where it did...

myedibleenso:
Looks like two tests are failing...

danebell:
The failing tests are:

>mTOR inhibitor Rapamycin

>HDAC inhibitors including trichostatin A completely restored RECK

Both tests check for a negative activation using the trigger 'inhibitor'. I think under the new heuristic as defined, these _should_ fail. They have the construction "X inhibitor" and are being interpreted as atomic Simple_chemicals, i.e. not subject to activation rules. @MihaiSurdeanu is this the desired behavior?

MihaiSurdeanu:
I think the tests are theoretically correct. In an ideal world, we should be able to peek inside the NEs to identify these sub-NE relations. Clearly, this is beyond our capability now. Plus, even if correct, I doubt these relations are that useful. I am Ok with commenting these 2 tests out.

PR #233, opened 2016-06-11
support controlled and controller regulations

marcovzla:
We were only considering nested regulations in the "controlled" slot, but
they can also nest as "controllers".

We encountered this problem on the sentence: "In the present study , we have further shown that ERK induced RhoA phosphorylation enhances the formation of actin stress fibers , which is consistent with the increased RhoA activity induced by ERK mediated RhoA phosphorylation ."


marcovzla:
test would be nice

MihaiSurdeanu:
Can you please add a test for that sentence?

danebell:
I am still working on the tests. Sorry for the premature commit. Also, I am editing DarpaActions to prefer complex controlleds over simple controlleds -- otherwise we might get inconsistent output, I think.

myedibleenso:
Why were we sorting the regs previously?  Anyone remember?  Was it something to do with the exporter?

marcovzla:
it was needed by mkRegulation, but I don't remember the details

marcovzla:
thanks!

PR #235, opened 2016-06-12
CSVParser -> DSVParser resolves #234

### LINKED ISSUE #234: CSVParser is broken
Running `PaperReader` against a `.csv` file doesn't produce any mentions.  I propose a succinct rewrite that handles both `.csv` and `.tsv` files.

myedibleenso:
This rewrite of the `CSVParser` (now `DSVParser`) produces valid FriesEntries and now supports both `.tsv` and `.csv` file types.  Accompanying these changes are two tests which specifically address `PMC1234335.csv`.


marcovzla:
Thanks!

PR #240, opened 2016-06-17
Issue236

hickst:
Fix GPP re-labeling bug in ReachSystem.

### DIFF
--- a/src/main/scala/edu/arizona/sista/reach/ReachSystem.scala
+++ b/src/main/scala/edu/arizona/sista/reach/ReachSystem.scala
@@ -49,7 +49,7 @@ class ReachSystem(
 
   def mkDoc(text: String, docId: String, chunkId: String = ""): Document = {
     val doc = processor.annotate(text, keepText = true)
-    val id = if (chunkId.isEmpty) docId else s"${docId}_${chunkId}"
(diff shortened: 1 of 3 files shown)


myedibleenso:
@hickst, it seems like `resolveDisplay` and `resolveDisplayForArguments` belong inside the display package.
I realize that these methods were already in ReachSystem, but could they now be moved here: https://github.com/clulab/reach/blob/master/src/main/scala/edu/arizona/sista/reach/display/package.scala

... or somewhere under `extern`?

hickst:
Perhaps they could but what's the motivation for doing so? Are we going to call them more than once? Are they generally useful from other places?

myedibleenso:
It just seems to pertain only to export (or maybe only a subset of export formats?)...

I thought `ReachSystem` was only supposed to extract grounded mentions from papers.  Is that not the case anymore?

Picking a label is only important to export, and thus seems more relevant to `ReachCLI`/the internals of a particular set of exporters.

myedibleenso:
Shouldn't it be up to each exporter to figure out what it needs to do with the mentions?

Anyone else want to weigh in on this?

hickst:
If displayLabel is only relevant to export then it should probably not be a property of the central data structure in the first place.

If you want to move this code, then file another issue but let's move on with this bug fix as we have lots of other more important things to do.

PR #242, opened 2016-06-23
hack to support nested regulations in FriesOutput

marcovzla:
We encountered a 2 level nested regulation in one of the dryrun papers.
Once again the serializer failed.

This is a hack to be able to continue with the evaluation.
A real solution would be to rewrite the serializer from scratch.


MihaiSurdeanu:
Thanks Marco

PR #247, opened 2016-07-09
Added rules for "VBN with" Regulations

myedibleenso:
I've added a pair of rules to cover regulations involving triggers of the form "VBN with" (ex. "stimulated with").


myedibleenso:
@MihaiSurdeanu, I had to modify your test by replacing "NRG" as the controller with "leptin", as `NRG` is being found as a cell line...

PR #255, opened 2016-07-13
Detect autophosphorylations (resolves #252)

### LINKED ISSUE #252: Extracting autophosphorylation
Some proteins autophosphorylate, that is, they phosphorylate themselves. This can happen in cis (cis-autophosphorylation) when a single molecule phosphorylates itself or in trans (trans-autophosphorylation) when one molecule phosphorylates another molecule of the same type while they are in a complex (typically a dimer). This is the case for EGFR which trans-autophosphorylates in a dimer.

Some simple examples:
- EGFR autophosphorylates on a tyrosine residue.
- EGFR phosphorylates itself. _(strictly speaking not correct but people sometimes say this)_

I'm guessing there are two approaches to extracting such events, either as a new modification type called "autophosphorylation" with a single EGFR argument or the same phosphorylation event as for 2 different entities but with EGFR phosphorylating EGFR. 

This comes up in PMC534114:
- "However, stimulation with EGF for varying time intervals revealed no significant differences in the levels of autophosphorylation of EGFR in cells expressing wild type Gab1 versus the Gab1 F446/472/589 mutant.", 
- "As has been previously reported, recruitment of Shp2 by Gab1 does not alter the magnitude or kinetics of tyrosine autophosphorylation of EGFR.", 
- "Levels of EGFR autophosphorylation are represented linearly following quantitation by densitometry and normalization for protein expression levels.", 
- "Because the substrates of Shp2 are for the most part unknown, we were additionally interested in examining the state of EGFR tyrosine phosphorylation following treatment with EGF in order to determine if the failure of Gab1 to bind p85, and potentially recruit Shp2, would influence levels of EGFR autophosphorylation.", 
- "The experiment presented in Fig shows that all cell lines exhibit EGFR autophosphorylation in response to EGF treatment, while only cells expressing the ectopically introduced ErbB3 protein show ErbB3 tyrosine phosphorylation in response to EGF stimulation."

myedibleenso:
Per @bgyori's request, I've added support for the detection of autophosphorylation events.  

# Changes
1. Tests for the examples in #252.
2. `AutoPhosphorylation` added to the taxonomy under `Phosphorylation`.  An `AutoPhosphorylation` event requires a "theme" and `cause`, which is transformed into a `Positive_regulation` where the `controller` matches the "theme" of the `controlled` (an `AutoPhosphorylation`).  This transformation basically mirrors how we currently handle `SimpleEvents` with a `cause`.
3. A new rule template for "auto" events (we may want to eventually add autoubiquitination, autoacetylation, etc.).  Currently, the template contains only four syntactic patterns, though I believe those four rules cover the majority of encountered structures.
4. A new action to properly handle these events. `splitSimpleEvents` could not be reused, as it disallows overlapping controllers and controlleds.

### DIFF
--- a/src/main/resources/org/clulab/reach/biogrammar/events/simple-event-auto_template.yml
+++ b/src/main/resources/org/clulab/reach/biogrammar/events/simple-event-auto_template.yml
@@ -1,0 +1,53 @@
+
+- name: ${ eventName }_syntax_1_noun
+  priority: ${ priority }
+  example: "EGFR autophosphorylation."
+  label: ${ label }
+  action: ${ actionFlow }
+  pattern: |
+    trigger = [lemma=/${ nominalTriggerLemma }/ & ${triggerPrefix} & tag=/^N/ & !mention=ModificationTrigger]
+    # these two should be the same
+    cause:BioChemicalEntity = nn
+    theme:BioChemicalEntity = nn
+    site:Site? = /prep_(at|on)|nn|conj_(and|or|nor)|cc/{,2}
+
+
+- name: ${ eventName }_syntax_2_noun
+  priority: ${ priority }
+  example: "The autophosphorylation of EGFR."
+  label: ${ label }
+  action: ${ actionFlow }
+  pattern: |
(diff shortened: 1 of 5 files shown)


hickst:
Nice cleanup, en passant, in DarpaActions

PR #256, opened 2016-07-14
Site pattern fix in event rule (resolves #254)

### LINKED ISSUE #254: Amino acid site missing when it appears before PTM
The tyrosine residue is missing from the extracted phosphorylation event (PMC534114): 
"We did not detect a change in the tyrosine phosphorylation of EGFR in cells expressing Gab1 proteins that are deficient in recruitment of Shp2."

When changed to: "We did not detect a change in the phosphorylation of **EGFR on tyrosine** in cells expressing Gab1 proteins that are deficient in recruitment of Shp2.",
tyrosine is recognized correctly.

myedibleenso:
I've added the example sentence @bgyori as a new test.

The fix for #254 ended up needing only one keystroke.  Wish they were all like this...


myedibleenso:
https://github.com/clulab/reach/blob/205708762d50a4496e503349d74f2a0968e8a970/src/main/resources/org/clulab/reach/biogrammar/events/simple-event_template.yml#L108

PR #265, opened 2016-07-19
Isolate flattening of controllers to export

### LINKED ISSUE #184: When the Controller of a reg is an event, convert it to a Controller entity + participant features:
This might be done already? 
Make sure it shows in the FRIES format.

myedibleenso:
These changes resolve #184, which is a major change to the system.

## Summary
* `reach` now produces nested events where `Event -> Entity` conversion (ex. `Binding -> Complex` when serving as the `controller` of a `ComplexEvent`) is postponed until export (see `OutputDegrader`).   I think this is a step in the right direction for `reach`.

* Actions associated with `ComplexEvents` have been simplified. 
* Changes to existing tests to reflect these changes, as well as new tests for the "flattening".

### DIFF
--- a/src/main/resources/org/clulab/reach/biogrammar/events_master.yml
+++ b/src/main/resources/org/clulab/reach/biogrammar/events_master.yml
@@ -243,6 +243,7 @@ rules:
       actionFlow: "mkBioMention"
       priority: "6"
 
+
   # Positive Regulation
   - import: org/clulab/reach/biogrammar/events/pos-reg_template.yml
     vars:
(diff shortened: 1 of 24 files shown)


MihaiSurdeanu:
Good job. convertEventToEntity is nice. 
I wonder if flattenController can reuse the code in convertEventToEntity instead of reimplementing almost the same functionality?

myedibleenso:
>I wonder if flattenController can reuse the code in convertEventToEntity instead of reimplementing almost the same functionality?
Ah, yeah I'd like to refactor that, but it didn't hit me immediately.  For now, I suggest we merge this.  One day someone can attempt to refactor that, but I think even so this is still more manageable that what we had.

PR #266, opened 2016-07-19
Refactor convertEventToEntity & flattenController

myedibleenso:
Refactored `convertEventToEntity` to generate either 
1) an event's output representation (ex. a `theme` or `controlled`) or 
2) its `cause` or `controller`.  Behavior is parameterized  (see`asOutput` boolean).


MihaiSurdeanu:
Nice.

Issue #210, opened 2016-05-25
Pass the whole paper to Reach rather than a collection of section


MihaiSurdeanu:

marcovzla:
closed by #267

PR #267, opened 2016-07-20
WIP don't split text document

marcovzla:
This pull request introduces a new nxmlreader that doesn't split papers into FriesEntries


marcovzla:
Why is travis not running?

marcovzla:
closes #210

myedibleenso:
WIP because you're unable to determine if we're reading the same things as before?

PR #273, opened 2016-07-21
Issue 257

enoriega:
Passes the translocation unit test and adds active voice for translocations

### DIFF
--- a/src/test/scala/org/clulab/reach/TestFeedback2.scala
+++ b/src/test/scala/org/clulab/reach/TestFeedback2.scala
@@ -56,8 +56,13 @@ class TestFeedback2 extends FlatSpec with Matchers {
   }
 
   val s7 = "We observed that endogenous ASPP2 translocates from cell/cell junctions to the cytosol/nucleus following RAS activation"
-  s7 should "contain 1 or more translocation events" in {
-    // TODO: we should pick 4 translocation events here - ENRIQUE
+  s7 should "contain 2 translocation events" in {
+    // TODO: we should pick 2 translocation events here - ENRIQUE
+    // ENRIQUE: Cell is in the black list of the NER, so there will be only two events
+    val mentions = getBioMentions(s7)
+
+    hasEventWithArguments("Translocation", List("ASPP2", "cell junctions", "cytosol"), mentions) should be (true)
+    hasEventWithArguments("Translocation", List("ASPP2", "cell junctions", "nucleus"), mentions) should be (true)
   }
 
   s7 should "contain 1 positive regulation" in {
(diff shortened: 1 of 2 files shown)


src/test/scala/org/clulab/reach/TestFeedback2.scala
```diff
@@ -56,8 +56,13 @@ class TestFeedback2 extends FlatSpec with Matchers {
   }
 
   val s7 = "We observed that endogenous ASPP2 translocates from cell/cell junctions to the cytosol/nucleus following RAS activation"
-  s7 should "contain 1 or more translocation events" in {
-    // TODO: we should pick 4 translocation events here - ENRIQUE
+  s7 should "contain 2 translocation events" in {
+    // TODO: we should pick 2 translocation events here - ENRIQUE
```
marcovzla: Is it ok to change the test's expected behavior? This was based on Mitre's feedback.
@MihaiSurdeanu opinions?

src/test/scala/org/clulab/reach/TestFeedback2.scala
```diff
@@ -56,8 +56,13 @@ class TestFeedback2 extends FlatSpec with Matchers {
...
+    // ENRIQUE: Cell is in the black list of the NER, so there will be only two events
```
marcovzla: Is it ok to change the test's expected behavior? This was based on Mitre's feedback.
@MihaiSurdeanu opinions?

myedibleenso: >ENRIQUE: Cell is in the black list of the NER, so there will be only two events

Is the black list not working correctly?

@enoriega, what are the four translocations that you're getting?

enoriega: It works correctly. There are two sources and two destinations, but one of the sources is `cell` which is filtered by the NER. So only two translocations match, where the source is `cell junctions` and the destinations are `cytosol` and `nucleus`

myedibleenso: What's with the 4 vs. 2 Translocations then?

myedibleenso: 4 before the blacklist?  2 with the blacklist?

enoriega: Yes

myedibleenso:
lgtm, but  I guess we can wait for Mihai to wake up and weigh in on this.

src/test/scala/org/clulab/reach/TestFeedback2.scala
MihaiSurdeanu: I agree with Enrique. Ok to change the test.

PR #275, opened 2016-07-22
Renamed PandasOutput to TSVOutput

enoriega:
Small cosmetic change. It is not being used anywhere and won’t impact any test of output


myedibleenso:
Why not just remove this from reach if it isn't being used anywhere?  It sounds like it was something you needed for context experiments.

enoriega:
Yes, we don't need it anymore. I will remove by tomorrow unless someone wants to keep it.

enoriega:
Won't proceed as this code file will be removed

Issue #274, opened 2016-07-22
Incorrect NER labeling of entities


hickst:
1) The proteins 'Smad 2', 'Smad 3', 'Smad 4', etc. are *already* in the Uniprot KB (in this form).
2) They appear in the ...kb/ner/Gene_or_gene_product.tsv.gz file (ex: 'SMAD 2')
3) They are not overridden in the NER override file.
4) The shorter entry 'SMAD' *does* appear in the NER override table as a family.

The number SMADs should be identified/labeled as GGP and later grounded with correct uniprot ID, but that is not happening which results in 'smad' labeled as a Family and '2' being ignored altogether.

I suspect that some component of NER resolution is not using the longest prefix string available.

MihaiSurdeanu:
@hickst: I can't replicate this issue in processors. See the last 2 tests in this file:
https://github.com/clulab/processors/blob/master/src/test/scala/org/clulab/processors/TestBioNLPProcessor.scala

They both pass. If you find a NER situation that fails related to this issue, can you please add a unit test for it in processors? If it's a grounding issue, can you please take over?

hickst:
Well, a Processors test is certainly a good first step. But since the NER spans Processors and Reach we should have the same tests in Reach, too. I've added your processor tests (and several others) to the Reach suite in the 'iss274' branch. These tests indicate that grounding contains the necessary entries and is working correctly when given 'smad 2'.

Note the failed result of your same Processors test at the Reach shell:
```
Welcome to ReachShell!

>>> Smad 2 is doing something.
sentence #0
Smad 2 is doing something .
Tokens: (0,Smad,NN), (1,2,CD), (2,is,VBZ), (3,doing,VBG), (4,something,NN), (5,.,.)
smad 2 be do something .
roots: 3
outgoing:
	0: (1,num)
	1:
	2:
	3: (0,nsubj) (2,aux) (4,dobj)
	4:
incoming:
	0: (3,nsubj)
	1: (0,num)
	2: (3,aux)
	3:
	4: (3,dobj)

entities:
mention text: Smad
List(Gene_or_gene_product, MacroMolecule, Equivalable, BioChemicalEntity, BioEntity, Entity, PossibleController)
	------------------------------
	Rule => ner-gene_or_gene_product-entities
	Type => CorefTextBoundMention
	------------------------------
	Protein|List(Gene_or_gene_product, MacroMolecule, Equivalable, BioChemicalEntity, BioEntity, Entity, PossibleController) => Smad
[...]

PR #281, opened 2016-07-26
Iss274

hickst:
Implements NER rules fix and tests for issue #274.


src/test/scala/org/clulab/reach/TestModelEntities.scala
```diff
@@ -40,15 +39,13 @@ class TestModelEntities extends FlatSpec with Matchers {
...
-  // "s2a entities" should "have GPP label" in {
-  //   val mentions = getBioMentions(s2a)
-  //   mentions.isEmpty should be (false)
-  //   printMentions(Try(mentions), true)      // DEBUGGING
-  //   mentions.size should be (4)
-  //   mentions.count(_ matches "Gene_or_gene_product") should be (4)
-  // }
+  "s2a entities" should "have GPP label" in {
+    val mentions = getBioMentions(s2a)
+    mentions.isEmpty should be (false)
```
marcovzla: I think tests like this should be written as `mentions shouldBe empty` instead.
This way, if they fail, they can display an error message that is more useful for debugging.

src/test/scala/org/clulab/reach/TestProteinResolutions.scala
```diff
@@ -59,6 +59,15 @@ class TestProteinResolutions extends FlatSpec with Matchers {
     (imkbP.resolve("stat1").isDefined) should be (true)
     (imkbP.resolve("STBA").isDefined) should be (true)
     (imkbP.resolve("stba").isDefined) should be (true)
+    (imkbP.resolve("SMAD2").isDefined) should be (true)
```
marcovzla: `imkbP.resolve("SMAD2") shouldBe defined` would provide a more informative error message.
Currently we would get something like "false is not true"

PR #287, opened 2016-07-27
handle n-ary bindings

### LINKED ISSUE #188: Add complexes_with relation
This should be done only for explicit mentions of complexes, e.g.: "the A-B-C complex"

marcovzla:
Added rule to handle n-ary bindings. Also added unittest. Closes #188


MihaiSurdeanu:
build fails.

marcovzla:
ooops

bgyori:
I noticed that currently a sentence like "The A-B-C complex activates D" would be extracted as 3 events with A-B, A-C and B-C as controllers. Is the goal to instead extract "A-B-C" as the controller? If that's the case, I think it's a great idea!

MihaiSurdeanu:
Yes, that's the goal... Thanks!

bgyori:
So this is merged into master now, right? I think I'm still getting the breakup into 2-way complexes in my latest build.

bgyori:
One sentence I tested just now is: `The ARAF/BRAF/RAF1 complex activates MAP2K1.` (this sentence is just for testing purposes, not necessarily correct biology). I get 3 `activation` frames and 3 `complex-assembly` frames for all 2-way complexes.

MihaiSurdeanu:
Ben is correct: the sentence above should be parsed as PosAct(Controller:Binding(ARAF, BRAF, RAF1), Controlled:MAP2K1). What this means is that we need to change the action for the rules the match "A/B/C/... complex" to allow for Binding events with more than 2 participants. To formalize:

1. We do NOT change the original Binding actions for the typical binding. That is, for the text "A binds B and C" continue to be translated into 3 binary Binding events, because this is the conservative interpretation.

2. BUT, for the texts such as "A/B/C complex" we know with 100% confidence that this is a complex with 3 participants. ONLY for such cases, the rule syntax and the corresponding action should allow for the output Binding event to have 3 participants.
Please let me know if this is not clear. If it is clear, please adjust the "A/B/C complex" rule+action to allow for this.
Thanks!

marcovzla:
Ok, I will get this done today.

I have a few questions related to the biomedical interpretation of this sentence.

In the sentence `The ARAF/BRAF/RAF1 complex activates MAP2K1.`, is the controller the complex or the binding event? are we sure that the complex was formed by a single binding event or could it be formed by several subsequent ones? Would the complex not trigger the activation if the patient gets it from a pill or an injection instead of a binding event?

bgyori:
> In the sentence The ARAF/BRAF/RAF1 complex activates MAP2K1., is the controller the complex or the binding event? 

The controller is the complex entity, not the binding event. The proteins, once they are in a complex, catalyze a given process. The event by which they got into a complex is not directly relevant. 

> are we sure that the complex was formed by a single binding event or could it be formed by several subsequent ones? 

Higher order complexes are always formed by elementary (2-way) binding events - at a low level it is not really possible for 3 molecules floating around to all bump into each other at the same time.
However, if all we know is that `A-B-C` are in a complex, we cannot assume that `A binds B`, `B binds C` and `A binds C` but only that `A can be in a complex with B`, etc. This is because it might be the case that `A` acts as a scaffold and both `B` and `C` bind to `A` only and not to each other. 

> Would the complex not trigger the activation if the patient gets it from a pill or an injection instead of a binding event?

It would trigger the activation just the same - doesn't really matter where the complex came from.

marcovzla:
Thanks @bgyori!

hickst:
> However, if all we know is that A-B-C are in a complex, we cannot assume that A binds B, B binds C and A binds C but only that A can be in a complex with B, etc. This is because it might be the case that A acts as a scaffold and both B and C bind to A only and not to each other. 

Isn't this exactly the description of the case for which MITRE asked that we use "complexes-with" and not "binding"?

marcovzla:
Yes it is. But it was my understanding that we should produce a Binding event internally and report complexes-with in the output. @MihaiSurdeanu is this correct?

MihaiSurdeanu:
Correct.

PR #303, opened 2016-08-02
Turn n-ary complex into n-ary binding event

### LINKED ISSUE #300: Turn n-ary complex into n-ary binding event
Convert mentions of `A-B-C complex` into a single binding event with three participants (without binarizing the event)

See discussion in #287

marcovzla:
This pull request implements a new action for binding events that doesn't splits them into pairs. This action is used by the `binding_token_5` rule to generate a single binding event with all the members of the complex.

Closes #300

### DIFF
--- a/src/main/scala/org/clulab/coref/CorefUtils.scala
+++ b/src/main/scala/org/clulab/coref/CorefUtils.scala
@@ -62,7 +62,7 @@ object CorefUtils {
   def argsComplete(args: Map[String,Seq[CorefMention]], lbls: Seq[String]): Boolean = {
     lbls match {
       case binding if lbls contains "Binding" =>
-        args.contains("theme") && args("theme").length == 2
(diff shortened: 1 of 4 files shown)


src/main/scala/org/clulab/coref/CorefUtils.scala
```diff
@@ -62,7 +62,7 @@ object CorefUtils {
   def argsComplete(args: Map[String,Seq[CorefMention]], lbls: Seq[String]): Boolean = {
     lbls match {
       case binding if lbls contains "Binding" =>
-        args.contains("theme") && args("theme").length == 2
+        args.contains("theme") && args("theme").length >= 2 // Binding is not required to be binary anymore (see binding_token_5)
```
marcovzla: @danebell I had to change this check to support bindings with more than two participants

marcovzla:
Note that this produces a Binding event, not the `complexes-with` relation expected during evaluation.

myedibleenso:
lgtm

PR #305, opened 2016-08-02
Issue259

### LINKED ISSUE #259: Past tense phosphorylation, inconsistent extractions
I am not sure what is going on here so I'll just post some observations:

From this sentence (PMC4331489)
`However, while MEK5D phosphorylated a kinase dead mutant of ERK5 (ERK5-KD) at its TEY site...`
2 events are extracted: a correct one that says MEK5D phosphorylates ERK5 and an incorrect one in which MEK5D gets phosphorylated.
 
Interestingly, `However, while MEK5D phosphorylated ERK5.` produces no events, `MEK5D phosphorylated ERK5.` produces no events but `MEK5D phosphorylated an ERK5.` produces a correct phosphorylation event between MEK5D and ERK5.

Further, `However, while MEK5D ubiquitinated a kinase dead ERK5.` produces the correct ubiquitination event but `However, while MEK5D phosphorylated a kinase dead ERK5.` again produces an incorrect phosphorylation event in which MEK5D is phosphorylated.

danebell:
These are slight adjustments to the templatic simple event rules to cover issue 259's problems. Although @myedibleenso prescribed a token-based rule to cover cases like "A phosphorylated B." in which (currently) the parser wrongly gives `phosphorylated` a tag of `VBN` (it should be `VBD`), I instead opted to remove the restriction on tags in the closest syntax rule. The reasons for this decision are that (1) the rule still requires correct dependency relations if not the correct tag, which *is* assigned here, so this is not a risky change, and (2) a token rule for this case hasn't been missed until now, leading me to think there is good coverage here and that introducing such a rule would introduce new false positives but few true positives.

Closes #259

### DIFF
--- a/src/main/resources/org/clulab/reach/biogrammar/events/simple-event_template.yml
+++ b/src/main/resources/org/clulab/reach/biogrammar/events/simple-event_template.yml
@@ -18,7 +18,7 @@
   label: ${ label }
   action: ${ actionFlow }
   pattern: |
-    trigger = [lemma=/${ verbalTriggerLemma }/ & ${triggerPrefix} & tag=/^(V|JJ)/]
+    trigger = [lemma=/${ verbalTriggerLemma }/ & ${triggerPrefix} & tag=/^(V|JJ)/ & !mention=ModificationTrigger]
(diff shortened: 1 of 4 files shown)


myedibleenso:
That's a better solution.  Thanks, @danebell.

danebell:
I broke some PTM tests with my changes. Changes coming soon.

Issue #306, opened 2016-08-02
produce `complexes-with` relation for complex mentions


marcovzla:
We are currently capturing complex mentions as binding events among its participants.
We need to represent this in the output as `complexes-with` relation

marcovzla:
Complexes were captured in #303

hickst:
Depending on how this is implemented, the outputter may need to be changed to accomodate this also so I'm adding myself, too.

MihaiSurdeanu:
There are two situations:
- A Binding event serves as Controller of an act/reg. In such situations, Gus's code should "flatten" this event into its output entity, i.e., the corresponding Complex entity. This applies to any kind of Binding, binary of not. Please ask @myedibleenso to handle this.

- A n-ary Binding event occurs standalone in the text, e.g., "We experimented with the A-B-C complex". In these cases, we do not do anything. We simply report an n-ary Binding. Hans promised to convert this into a complexes_with relation in the index card format. @marcovzla, please email Hans about this situation, with an example (even if artificial), and tell him about the needed conversion. Cc me and Tom please.

hickst:
Send sample output to Hans and Cheryl.

myedibleenso:
>A Binding event serves as Controller of an act/reg. In such situations, Gus's code should "flatten" this event into its output entity, i.e., the corresponding Complex entity. This applies to any kind of Binding, binary of not. Please ask @myedibleenso to handle this.

This doesn't really sound flattened, but I can do it.  We just throw away the trigger and change the label from `Binding` to `Complex`, right?

MihaiSurdeanu:
Yes. But I think the fundamental difference there is the Complex should be
an entity (it's the output of the event), whereas Binding is the event
itself. In the end this is semantics though. As long as you can uniquely
identify these for Hans, we're good.

marcovzla:
What about a Binding that was split into binary bindings? should we attempt to put it back together? or just create several binary complexes?

myedibleenso:
@marcovzla, it might be easier if you give him an example to look at.

myedibleenso:
I'm guessing no, though.

MihaiSurdeanu:
Just create several binary complexes. Remember that we only do this if they
serve as Controllers for act/reg. In such situations, I think the text is
more likely to state something like "the A-B... complex regulates
something" than "the binding between A, B, and C regulates something".

myedibleenso:
Yeah, this is already being done.  Will put in a PR for the tests.

[...]

PR #307, opened 2016-08-03
Added more tests for degradation of output

myedibleenso:
A few tests to ensure that a `Binding` serving as a `controller` of some event is converted into a `Complex`, while a `Binding` event on its own (i.e., not serving at another event's `controller`) left untouched.


myedibleenso:
Related to #306 (not sure if anything else is needed to close that one).

myedibleenso:
Thanks, @hickst.  That was sloppy of me...

PR #311, opened 2016-08-04
solves the "key not found: controller" encountered in the 1K

myedibleenso:
I ran this on the four papers that failed with "key not found: controller" in the 1K run.  They all pass.


hickst:
> the purpose of this method *seems* to be to filter out duplicates. Is that the case?
> we can do that pretty easily in assembly...maybe this could be retired/rewritten?

But there are still some use cases for running without Assembly so I don't think we want to move functionality exclusively to Assembly. Especially, given how much faster reading is without assembly, at the moment.

PR #315, opened 2016-08-10
Rewrite of FriesOutput

hickst:
This rewrite fixes the 13 FRIES-output errors of the 1K test run. All DR3 files also succeed.


src/main/scala/org/clulab/reach/extern/export/fries/FriesOutput.scala
```diff
@@ -198,115 +196,119 @@ class FriesOutput extends JsonOutputter {
...
+                               otherMetaData: Map[String, String]): PropMap = {
+    val model:PropMap = new PropMap
+    addMetaInfo(model, paperId, startTime, endTime, otherMetaData)
+
+    val frames = new FrameList
+    model("frames") = frames
+
+    for (mention <- mentions) {
       mention match {
         case em:BioTextBoundMention =>
```
myedibleenso: Forgive my ignorance, but is this method meant to handle any kind of Entity?  Does the outputter consider a complex to be an entity?  If so, maybe you want to match on the Mention's label, rather than the type of the Mention.

src/main/scala/org/clulab/reach/extern/export/fries/FriesOutput.scala
```diff
@@ -321,205 +323,209 @@ class FriesOutput extends JsonOutputter {
...
+  /** Lookup the unique ID, for the given mention, in the given maps, based on the type
+      of the mention. Returns an option for the ID found, or None if no ID found. */
   private def lookupMentionId (mention: Mention,
-                               entityMap:IDed,
-                               eventMap:IDed): Option[String] = mention match {
+                               entityMap: IDed,
+                               eventMap: IDed): Option[String] = mention match {
     case tbm: TextBoundMention => entityMap.get(tbm)
     case evm: EventMention => eventMap.get(evm)
+    case rem: RelationMention => eventMap.get(rem)
```
myedibleenso: Wouldn't something like `case m: Mention => entityMap.getOrElse(m, buildMePlease(m))` work for all three cases?

src/main/scala/org/clulab/reach/extern/export/fries/FriesOutput.scala
```diff
@@ -321,205 +323,209 @@ class FriesOutput extends JsonOutputter {
...
-          for ((p, i) <- ms.zipWithIndex) {
-            assert(p.isInstanceOf[TextBoundMention], s"complex participant is not an instance of TextBoundMention: ${p}")
-            if (!entityMap.contains(p)) {
-              throw new RuntimeException(s"Complex participant [${p.text} [mods: ${p.toCorefMention.modifications.map(_.toString).mkString(" ")}}]] not in entityMap \nin event [$currEvent] \nin sentence[${p.document.sentences(p.sentence).words.mkString(" ")}]:\n" + p.json(pretty = true))
-            }
+        val complexArgs = arg.asInstanceOf[RelationMention].arguments
+        for (key <- complexArgs.keys.toList.sorted) {
+          val argSeq = complexArgs.get(key).get
+          for ((p, i) <- argSeq.zipWithIndex) {
+            if (!isTextBoundMention(p))
```
myedibleenso: I guess confirming that the participant matches the label `Entity` isn't enough here, huh?  I'm just wondering about something like a complex of a complex.  We're "flattening" everything now, but who knows what we'll be doing in the future.

src/main/scala/org/clulab/reach/extern/export/fries/FriesOutput.scala
```diff
@@ -321,205 +323,209 @@ class FriesOutput extends JsonOutputter {
...
+  /** Return a tuple of maps mapping entities to unique IDs and events to unique IDs. */
+  private def makeEMaps (paperId: String,
+                         mentions: Seq[Mention],
+                         passageMap: Map[String, FriesEntry]): (IDed, IDed) = {
+    val entityMap = new IDed
+    val eventMap = new IDed
+
+    for (mention <- mentions) {
+      mention match {
+        case em:BioTextBoundMention =>
```
myedibleenso: I feel like labels would be more future-proof (ex. `case entity if entity matches "Entity" =>`), but if it works...

myedibleenso:
Glad to hear you were able to find solutions for those errors, @hickst.  Thanks so much for taking care of it.

PR #329, opened 2016-08-24
Add hoc solution to text preprocessing for the nxml indexer

enoriega:
This is a temporary solution while Marco comes back from vacation.


MihaiSurdeanu:
Can you please add this before line 55: 
// This is potentially incorrect because this preprocesses both text and NXML tags...
// TODO: this needs to be fixed by adding a preprocessing callback to NxmlReader

PR #345, opened 2016-09-03
Plain text reader

myedibleenso:
These changes add support for processing plain text files.  The expected file encoding is read from `application.conf` (see `encoding`).


johnbachman:
Does this mean that plain text files can be read without first breaking up the contents into .tsv/.csv files as before?

MihaiSurdeanu:
Yep

PR #361, opened 2016-09-13
org.clulab.assembly -> org.clulab.reach.assembly

myedibleenso:
This is simply a reorganization of packages.  Everything in `assembly` relies on the output of `reach`, so it seems to make more sense to place `assembly` under `reach`.


myedibleenso:
Ah, nevermind.  I just realized that the precedence classifier will need to be retrained in order to be compatible with this structure.

PR #362, opened 2016-09-14
Problems with complex formation from ligation

### LINKED ISSUE #338: Problems with complex formation from ligation
We encountered a few suspicious extractions of complex formations based on the occurrence of the word `ligation`. 

Two sentences where this seems to be the issue are:
- It has been reported that there is increased activity of MAPKs with ligation of ICAM-1, VCAM-1 or E-selectin ligation in epithelial and endothelial cells.
- Rac GTPases are activated by a broad array of extracellular stimuli relevant to RA, including chemokines, lymphocyte antigen receptor ligation, inflammatory cytokines, and cell-cell adhesion

from which `ICAM-1 binds VCAM-1` and `antigen binds antigen receptor` are extracted, respectively.

I tried a few synthetic sentences and I found that `antigen receptor ligation` in itself doesn't result in any events being extracted but `lymphocyte antigen receptor ligation` produces the previously seen `antigen binds antigen receptor` event. 

The rule being implicated is `binding7a`.

marcovzla:
This pull request fixes #338 by:
- disallowing overlapping binding participants
- removing the word "ligation" from the binding triggers

Note that "ligate" used as a verb was not removed. See for example rule "binding_16"
https://github.com/clulab/reach/blob/0206ec288cd1fae1372b21e8c5cd01d0a19959a0/src/main/resources/org/clulab/reach/biogrammar/events/bind_events.yml#L177-L184

### DIFF
--- a/src/main/resources/org/clulab/reach/biogrammar/events/bind_events.yml
+++ b/src/main/resources/org/clulab/reach/biogrammar/events/bind_events.yml
@@ -41,7 +41,7 @@
   action: mkBinding
   priority: ${ priority }
   pattern: |
-    trigger = [word=/(?i)binding|dimerization|heterodimerization|ligation|recruitment/] (?! [lemma="site"])
+    trigger = [word=/(?i)binding|dimerization|heterodimerization|recruitment/] (?! [lemma="site"])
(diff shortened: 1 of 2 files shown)


marcovzla:
Because this grammar is imported only once, so we didn't need to use variables.
Also, not all the rules use the same triggers.
However, we could use variables in this file for consistency.

PR #373, opened 2016-09-19
update processors to 6.0.0

marcovzla:


myedibleenso:
He switched to using an `Int` to specify `withDiscourse` for some reason: https://github.com/clulab/reach/compare/proc600...processors6

PR #377, opened 2016-09-26
Integrate logging into Reach

terroni:
These changes continue what was started in https://github.com/clulab/reach/pull/318 which setup the necessary configuration for the logger. I've simply searched for all print statements and verbose/debug flags and replaced them with the appropriate level (`logger.info`, `logger.debug`,  or `logger.error`). I imagine this should not affect functionality of any of these files, but I'm not familiar enough with the code base to know exactly what each file does so this could use another set of eyes. There were a few instances where I felt replacing a print statement was not suitable either because it just added a newline to the output or was connected to a PrintWriter.


hickst:
Terron,

We need another round on this before integration.

1) ReachShell is an interactive program: it should not output to login or the user will never see it.
2) Any println inside an exception should be level ERROR (example: `src/main/scala/org/clulab/reach/brat/Brat.scala`)
3) Most of the outputs in ReachCLI need to be level INFO (not DEBUG) -- this is a top-level program and we always need to see what it is doing as it is doing it (but we don't want the detail of the DEBUG level).
4) Current printlns whose message says "Warning" should probably be level WARN not DEBUG (example:  `src/main/scala/org/clulab/reach/context/ContextEngine.scala`)
5) Maybe others can comment on this but I think that
`src/main/scala/org/clulab/reach/display/package.scala`  was made to explicitly output text so it shouldn't be outputting to log file at all.

src/main/scala/org/clulab/reach/ReachShell.scala
```diff
@@ -1,16 +1,19 @@
...
+
 import scala.collection.immutable.ListMap
 import org.clulab.reach.display._
 import org.clulab.reach.context.ContextEngineFactory.Engine
 import RuleReader._
+import com.typesafe.scalalogging.LazyLogging
 
-object ReachShell extends App {
-  println("Loading ReachSystem ...")
+object ReachShell extends App with LazyLogging {
```
myedibleenso: Sorry about not being more specific when we discussed this issue, but the `ReachShell` should be using `println`, not `logger.info`.  Can you please revert your changes to this file?

src/main/scala/org/clulab/reach/brat/Brat.scala
```diff
@@ -35,7 +36,7 @@ object Brat {
           Some(TextBound(id, label, spans, chunks(2)))
         } catch {
           case e: Exception =>
-            println(line.trim)
+            logger.debug(line.trim)
```
myedibleenso: @hickst mentioned a different level here.

src/main/scala/org/clulab/reach/context/ContextEngine.scala
```diff
@@ -66,30 +67,30 @@ object ContextEngine {
   def getDescription(mention:BioMention, voc:Map[(String, String), String]):String = {
     val key = getContextKey(mention)
     if(key._2.startsWith("uaz:")){
-      println(s"Warning: ${mention.text}")
+      logger.debug(s"Warning: ${mention.text}")
```
myedibleenso: I believe these should also be a different level.  Please coordinate with @hickst on these.

src/main/scala/org/clulab/reach/display/package.scala
```diff
@@ -1,12 +1,14 @@
...
+import com.typesafe.scalalogging.LazyLogging
 import org.clulab.odin._
 import org.clulab.reach.mentions._
 import org.clulab.processors.{Document, Sentence}
+
 import scala.util.matching.Regex
 
 
-package object display {
+package object display extends LazyLogging {
```
myedibleenso: Sorry about not being more specific when we discussed this issue, but the `display` package should be using `println`, not `logger.info`.  Can you please revert your changes to this file?

src/main/scala/org/clulab/coref/Coref.scala
```diff
@@ -347,11 +347,9 @@ class Coref {
 
       if (mentions.isEmpty) Nil
       else {
-        if (debug) {
-          println("BEFORE COREF")
-          displayMentions(mentions, mentions.head.document)
-          println("Starting coref...")
-        }
+        logger.debug("BEFORE COREF")
+        displayMentions(mentions, mentions.head.document)
```
marcovzla: Is this now being called all the time?
`displayMentions` uses `println` to display the mentions.
We should use some other option that returns the string to print and sent it to the logger, so that we can turn it off.
Currently we are printing more than 4mb of stuff, and travis doesn't like that.

myedibleenso: The method that returns a string is `display.summarizeMentions` (same signature).  Maybe this could be changed to something like `logger.debug(display.summarizeMentions(mentions, mentions.head.document))`?

myedibleenso:
One more thing, @terroni...this is not a logging issue, but since you have a PR open can you please set this property to `false`?

https://github.com/clulab/reach/blob/master/src/main/resources/application.conf#L17

hickst:
@terroni Please change the location of the log file (specified in `application.conf`) to be consistent with all the other default locations: that is `${HOME}/Documents/reach` and please *do not* make the logfile hidden: the name `reach.log` is sufficient.

myedibleenso:
@marcovzla noted that we're printing tons of stuff that we don't need to be printing.  Minimally, the logging/debugging in `TestPaperReader` and `TestTranslocationEvents` needs to be silenced.

myedibleenso:
And this line, @terroni: https://github.com/clulab/reach/pull/377/files#diff-a73606c7414d02984238c2851c5c24a9R351

myedibleenso:
@danebell, have you had a chance to look at the `Coref` `foreach` stuff?

Issue #385, opened 2016-09-29
Optimize the rule based context engines' performance


enoriega:
Make the rule based context engines work faster.

myedibleenso:
Should #390 have closed this?

enoriega:
Yes

PR #387, opened 2016-10-01
Changed underlying data structures to optimize performance

enoriega:
Reduced running time by ~10%


MihaiSurdeanu:
@enoriega: I'm afraid this still needs some work. The featureVocabularyKeys and latentVocabularyKeys create a new array on the fly every time they are called. I suggest you sit down with @marcovzla or @myedibleenso and go through the code to clean it up.

myedibleenso:
Was this superseded by #390?  If so, can it be closed?

enoriega:
This is no longer needed after addressing Issue #385

PR #388, opened 2016-10-04
Save all grounding resolutions in the FRIES output

hickst:
The alternate grounding candidates are now output in the FRIES output in a JSON list titled 'alt-xrefs' (since the preferred grounding uses 'xrefs'). In order to make sense of and to use the alternate grounding candidates, however, it was also necessary to output the species in groundings, so this was added, too.


MihaiSurdeanu:
@bgyori @johnbachman: we now output all valid grounding options under "alt-xrefs".

bgyori:
Great, thank you!

PR #391, opened 2016-10-06
Breakup

myedibleenso:
We've broken reach up into three sub-projects à la `processors`:
1) `main` (coref, context, grounding, mention stuff, etc.
2) `export` (`JSONOutputter`, `ReachCLI`, etc)
3) `assembly` (the prodigal son has returned)

## What depends on what?
`export` depends on `main` and `assembly`
`assembly` depends on `main`


enoriega:
Are these three new projects going to be part of the same repository? The reason I ask this is because we have a context branch with code specific to research that is not part of `master` and plugs into `ReachCLI`. Maintaining multiple `context` branches in different repositories introduces difficulties for maintenance, additional to the potential burden of fixing the code in the specific research branch that gets broken as a result of a drastic refactoring

myedibleenso:
As I mentioned over slack, @enoriega, these three sub-projects will live in the same repo (see the [`breakup` branch](https://github.com/clulab/reach/tree/cfeefe8122d08793ca8d069841969aa7d439d091)).

enoriega:
Thanks @myedibleenso

myedibleenso:
Ugh...the tests for the sub-projects are being run in parallel (3 `BioNLPProcessors` are being instantiated simultaneously).

hickst:
I'm seeing a warning about a circular dependency:

`[warn] circular dependency found: org.clulab#reach_2.11;1.3.3-SNAPSHOT->org.clulab#reach-assembly_2.11;0.0.1->org.clulab#reach_2.11;1.3.2`

Is this something you know about? Is it going away when the breakup is complete?

marcovzla:
@hickst yes that was one of the motivations for doing this

myedibleenso:
@MihaiSurdeanu, have you had a chance to look this over?

PR #393, opened 2016-10-07
NxmlSearcher now save publication year and doc size.

MihaiSurdeanu:


MihaiSurdeanu:
Very simple change to NxmlSearcher to save publication year and doc size.
Assigned to you, @myedibleenso, because I'm not sure how this interacts with the code reorganization.

PR #396, opened 2016-10-11
Tweaks to arizona output format

myedibleenso:
## Modifications to "arizona" output format
- Remove non-event rows
- Resolve displayed text of event arguments


MihaiSurdeanu:
Excellent. Thanks!

PR #398, opened 2016-10-13
moved executables

marcovzla:
Moved executables to the root project and renamed it `reach-exe`.
Jars for root project are now generated when publishing and releasing (they were deactivated previously).


marcovzla:
We should probably rename `application.conf` to `reference.conf` in the subprojects,
and add an `application.conf` to `reach-exe`.

Do we want to move the API related stuff to `reach-exe` too?

MihaiSurdeanu:
Yes on both

myedibleenso:
`reach-assembly` now uses `reference.conf`: https://github.com/clulab/reach/blob/76b06926cf62520960cc08d2633598621e074eef/assembly/src/main/resources/reference.conf

PR #399, opened 2016-10-14
Corrections to Arizona output

myedibleenso:
- Activations with negative polarity have activated inputs that are deactivated at
output

- ~~ComplexEvents with negative polarity have a .neg suffix added to their controllers~~
- Polarity of a ComplexEvent is encoded in its label


myedibleenso:
Example output: https://gist.github.com/myedibleenso/9cee738263603de82bc09c383423c065

Issue #328, opened 2016-08-24
Confounding events affecting protein _amount_ with protein _activity_


bgyori:
This is more systemic issue that I wanted to bring up. In cell biology the process in which one protein regulates the _amount_ of another protein and the process in which one protein regulates the _activity_ of another protein are different and should therefore be differentiated in REACH extractions. The former, that is, the regulation of the _amount_ of a protein can happen via transcriptional regulation or, for instance, the regulation of protein degradation. The latter, that is, the regulation of the _activity_ of a protein typically happens via post-translational modification, or, for instance, translocation.

Consider the sentence `Ets-1 upregulates MMP-9 expression triggered by TGF-beta1 and TPA via MAPK signaling`. Here Ets-1 is understood to affect the _amount_ of MMP-9 protein, not its _activity_. Currently REACH extracts that `Ets-1 positively activates MMP-9`, which is not correct.

To solve this problem, I believe, REACH would need to be extended with events describing the expression of proteins and then allow regulations on this process.

bgyori:
Another example of the same problem: `Therefore, we hypothesize that loss of neurofibromin in monocytes and macrophages enhances ROS production [...]. `. Here ROS refers to reactive oxygen species (not proteins) whose _amounts_ are important for the cell. The event extracted should reflect that neurofibromin increases the amount of ROS, but instead, currently, neurofibromin is extracted as a positive activator of ROS.

MihaiSurdeanu:
Solution:
1. Add a simple event, Amount or Expression.
2. This is modeled as a Regulation, just like any other simple event. It should just work.
3. Modify the Fries format to allow for the new event.
4. Profit.

MihaiSurdeanu:
Dane, Gus - add the new simple event.
Tom - output

MihaiSurdeanu:
@bgyori: what triggers should we be using for detecting that it's about amounts: "expression", "production". What else?

johnbachman:
"level(s) of"

bgyori:
+ amount of, transcription of

danebell:
Should this event type be merged with [the currently inactive type `Transcription`](https://github.com/clulab/reach/blob/master/src/main/resources/org/clulab/reach/biogrammar/inactive/transcription_events.yml)?

johnbachman:
No, because there are many post-transcriptional ways of regulating protein abundance (mRNA stability, microRNAs, translation efficiency, protein stability, etc.).

johnbachman:
[...]

PR #401, opened 2016-10-14
Issue 328

danebell:
The goal of issue #328 was, broadly, to reintroduce Transcriptions/Expressions and to introduce the related Event types Amount, IncreaseAmount (of which Transcription is a hyponym), and DecreaseAmount. Tests are provided, and exporting and assembly have been altered to allow these events.


MihaiSurdeanu:
@danebell: 1 unit test failed here.

myedibleenso:
@danebell, I seem to have "fixed" the index card error.  

I think the index card tests should be disabled.  My impression is that if we ever need to produce index cards again, more work will be needed for them to validate.

hickst:
@myedibleenso I agree: the Indexcard output is way out-of-date and we should "mothball" this code until (IF?) it is ever needed again. The simplest way might be to just comment out its unit tests. We would still update it for major structural changes (i.e. to keep it compiling) but no longer "enhance" it to compete with the FRIES output.

p.s. thanks for fixing the IC output for this change.

MihaiSurdeanu:
@danebell, @myedibleenso: we now fail a bunch of activations, which should be regulations now due to the new Expression events?

myedibleenso:
>which should be regulations now due to the new Expression events?

Probably.  I guess those tests should be adjusted and moved to the Regs suite.  Maybe the Index card test(s) should be commented out as well?

MihaiSurdeanu:
Yep...

PR #403, opened 2016-10-17
Web service

myedibleenso:
example:

```scala
sbt "run-main org.clulab.reach.export.server.FileProcessorWebUI"
```

open your browser to [localhost:8888](http://localhost:8888)

### DIFF
--- a/assembly/src/main/scala/org/clulab/reach/assembly/export/AssemblyExporter.scala
+++ b/assembly/src/main/scala/org/clulab/reach/assembly/export/AssemblyExporter.scala
@@ -292,13 +292,23 @@ class AssemblyExporter(val manager: AssemblyManager) extends LazyLogging {
       event.predecessors.map(se => EERLUT(se.equivalenceHash))
   }
 
-  // FIXME: take columns and sep as params
   def writeRows(
     f: File,
     cols: Seq[String],
     sep: String = AssemblyExporter.SEP,
     rowFilter: Set[Row] => Set[Row]
   ): Unit = {
+
+    val results = rowsToString(cols, sep, rowFilter)
+    // write the output to disk
+    FileUtils.writeStringToFile(f, results)
+  }
+
+  def rowsToString(
+    cols: Seq[String],
+    sep: String = AssemblyExporter.SEP,
+    rowFilter: Set[Row] => Set[Row]
+  ): String = {
     val rowsForOutput = rowFilter(getRows)
     // validate output
     validateOutput(rowsForOutput)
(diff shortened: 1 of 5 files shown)


MihaiSurdeanu:
Nice!

hickst:
Doesn't work for me: HTTP method not allowed, supported methods: GET

Also why did all the API stuff get moved to a separate subproject and then this new server get placed in the previous API path in the main project? Surely this is as much an export as the API...

myedibleenso:
>Doesn't work for me: HTTP method not allowed, supported methods: GET

What browser did you use?  Did you use the UI, or did you attempt to use postman, curl, etc.?

>Also why did all the API stuff get moved to a separate subproject and then this new server get placed in the previous API path in the main project? Surely this is as much an export as the API...

I was told we wanted to put the runnables (at least some of them) in this top-level package.  I think your api stuff was also supposed to be moved to the`reache-exe` thing. @marcovzla?  Feel free to move it, @hickst...

marcovzla:
I didn't move the apis because they were not executables. But we can move them if that makes more sense.

hickst:
@myedibleenso I used Firefox. I haven't tried Postman yet but I suspect it might work (if I enforced a GET call) since the message seems to indicate that POST was used in the call.

@marcovzla No, if "executable" is the criteria, it makes sense to leave the APIs were they are.

hickst:
@myedibleenso Many form interfaces use POST by default. Perhaps the one you are using (Twist?) is doing this, too?

myedibleenso:
It's a POST.  Not using Twist, either.  Try it on another browser?

MihaiSurdeanu:
Yeah, I think the APIs should be in the main project.

myedibleenso:
So what is supposed to be in `reach-exe`?  I don't understand the separation anymore.

MihaiSurdeanu:
- export should be just the export code.
- exe: APIs for the whole shebangs + runnables in any form (cmd line, web)

hickst:
Oops, user error (sort of). I used a file from my temp area which turned out to be a mis-labeled HTML version of the NXML. It works fine when I use a real NXML file.

myedibleenso:
>Oops, user error (sort of). I used a file from my temp area which turned out to be a mis-labeled HTML version of the NXML. It works fine when I use a real NXML file.

That's funny to hear it's working for you.  It seems to work fine on Chrome and Safari, but I ran into trouble with Firefox as well when I tried that browser (didn't know it was a problem until you mentioned it).

hickst:
I was testing under another browser (Safari), as you suggested but the latest version works under Firefox, too.

PR #409, opened 2016-10-18
Create Gordo (fat jar) without running tests

hickst:
Assembly task appears to be including the test classes and running all the tests before it builds the Uber JAR. Since we will *never* create an Uber JAR without having thoroughly tested the code :), I added a line to the build file to prevent testing within the assembly task.


marcovzla:
LGTM

PR #418, opened 2016-10-25
Issue 397: Promoting aliases to mentions

### LINKED ISSUE #397: NER: Handle appositives for detecting abbreviations/acronyms; propagate globally

danebell:
In a case such as the following made-up sentences, one entity name `23peM` is not captured by our named entity recognizer. 
> We examine the role of 23peM. 23peM (usually called ASPP1) is a common protein.

Through the use of alias rules, though, we can detect that it ought to be. After these commits, such entities are recognized through the use of the Alias grammar followed by a custom HashTrie to capture the previously unrecognized entities. An alias found in one document will be added to this search in other documents it is processed with. This means that an alias relationship found in one section of a paper can be used to inform NER in the other sections.

This closes #397


MihaiSurdeanu:
@danebell: this is awesome! But some tests are failing...

myedibleenso:
It looks like there are three tests failing, but one of those tests appears to be a duplicate...

myedibleenso:
What do `aCapture` and `bCapture` represent?  Is `aCapture` the alias?

danebell:
One is a captured entity, and one its uncaptured alias, but they could be in either order, so `aCapture` is just the first text string and `bCapture` the second.

PR #437, opened 2016-11-04
deleted test

marcovzla:
This test is wrong and it is redundant with tests in processors.
I deleted this one but maybe the whole file should be removed?
The corresponding tests in processors are here: https://github.com/clulab/processors/blob/master/main/src/test/scala/org/clulab/odin/TestTokenPattern.scala


marcovzla:
I deleted the file since there is a more complete version of these tests in processors.

MihaiSurdeanu:
@marcovzla: So it's safe to update to 6.0.1 now?

marcovzla:
We need to wait until 6.0.1 is available on maven.
It is listed but it gives a 404 for some reason.
I opened an issue here: https://issues.sonatype.org/browse/MVNCENTRAL-1349

MihaiSurdeanu:
@marcovzla, @hickst, @myedibleenso: any updates on this? If not, should we release processors 6.0.2 today?

marcovzla:
The issue was resolved just now. We need to wait around an hour for the artifacts to propagate.

PR #439, opened 2016-11-05
Trigger inclusion in "arizona" output

### LINKED ISSUE #435: Include event triggers in tabular output
Based on today's conversation with CMU, we'll add a new column to the output that lists the lemmatized triggers for the evidence associated with each row.  We'll sort these event triggers by their frequency.  This will be used for analysis of regulations with the aim of eventually developing a subtype of regulations that model increase/decrease effects.

myedibleenso:
These changes resolve #435.  The elements of the`Triggers` column are lemmatized and sorted by frequency.

An example of the changes to the output format can be found here: https://gist.github.com/myedibleenso/2da0a2f101e89712115e116c4a98a6d4#file-pmc16533949-arizona-output-tsv

### DIFF
--- a/assembly/build.sbt
+++ b/assembly/build.sbt
@@ -6,7 +6,7 @@ libraryDependencies ++= {
   val akkaV = "2.4.3"
   Seq(
     "org.scalatest" %% "scalatest" % "2.2.4" % "test",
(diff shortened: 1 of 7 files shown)


MihaiSurdeanu:
Thanks!

PR #457, opened 2016-11-28
Cmuformat

MihaiSurdeanu:


myedibleenso:
I wonder if I should add `Source` and `Destination` as some kind of modification when converting a Translocation mention into a `SimpleEvent` `EER`.

assembly/src/main/scala/org/clulab/reach/assembly/export/AssemblyExporter.scala
```diff
@@ -423,6 +468,30 @@ object AssemblyExporter {
   val CONTEXT_CELLULAR_COMPONENT = "CONTEXT (CELLULAR COMPONENT)"
   val CONTEXT_TISSUE_TYPE = "CONTEXT (TISSUE TYPE)"
   val TRIGGERS = "TRIGGERS"
+  val TRANSLOCATION_SOURCE = "TRANSLOCATION (SOURCE)"
+  val TRANSLOCATION_DESTINATION = "TRANSLOCATION (DESTINATION)"
+
+  // columns for the CMU tabular format
```
myedibleenso: Maybe these should be under export/src/main/scala/org/clulab/reach/export/cmu/CMUExporter.scala?

MihaiSurdeanu: These are used in the "arizona" format.

assembly/src/main/scala/org/clulab/reach/assembly/export/AssemblyExporter.scala
```diff
@@ -451,6 +523,29 @@ object AssemblyExporter {
     AssemblyExporter.SEEN_IN
   )
 
+  val CMU_COLUMNS = Seq(
```
myedibleenso: ? `CMU_COLUMNS` doesn't appear to be used in `AssemblyExporter`...

MihaiSurdeanu: They will be soon. Working on it.

PR #476, opened 2017-01-14
Produce multiple output formats per input file

### LINKED ISSUE #475: Multiple output formats
Let's be able to output more than one output format at once. This is to be able to have multiple formats when doing big runs, like one over the whole open access subset

myedibleenso:
This closes #475.

Note that `sbt "runMain org.clulab.reach.ReachCLI"` is now `sbt "runMain org.clulab.reach.RunReachCLI"`


MihaiSurdeanu:
I think the default config should be all outputs: fries, arizona, cmu

PR #479, opened 2017-01-26
find shortest path between heads only

### LINKED ISSUE #426: REACH taking long time on certain texts
I'm not specifically interested in this abstract but it is one example on which REACH seems to take a very long time. I'm putting it in as an issue because looking at why it's so slow might expose some more general issue. If not, feel free to close this!

https://www.ncbi.nlm.nih.gov/pubmed/27551758

```
The effects of tanshinone IIA on hypoxia/reoxygenation-induced myocardial microvascular endothelial cell apoptosis in rats via the JAK2/STAT3 signaling pathway. This study aims to investigate the effects of tanshinone IIA on hypoxia/reoxygenation (H/R)-induced myocardial microvascular endothelial cell (MMEC) apoptosis in rats. MMECs from 10-days aged rats were isolated, cultured and identified, which were then divided into following groups: control group, control+tanshinone IIA (50\u03bcM) group, H/R model group, H/R+tanshinone IIA (5\u03bcM) group, H/R+tanshinone IIA (50\u03bcM) pre-treatment group, H/R+AG490 (50\u03bcM) pre-treatment group and H/R+AG490 (50\u03bcM)+tanshinone IIA (50\u03bcM) pre-treatment group. MTT assay, TUNEL staining and flow cytometry were used to measure the cellular viability and apoptosis. Western-blot were performed to detect protein expressions in JAK2/STAT3 signaling pathway. Compared with control group, H/R group showed decreased cell viability, increased apoptosis rate, increased proportions of cells into G0/G1 phase, decreased proportions of cells in S phase and G2/M phase, as well as up-regulated expressions of JAK2, STAT3, p53, Bax, Caspase-3, pJAK2 and pSTAT3, and down-regulated Bcl-2 expression (all P<0.05). Compared with H/R group, H/R+tanshinone IIA (5\u03bcM) group, H/R+tanshinone IIA (50\u03bcM) group H/R+AG490 (50\u03bcM) group and H/R+AG490 (50\u03bcM)+tanshinone IIA (50\u03bcM) group had increased cell viability, decreased apoptosis rate, reduced proportions of cells into G0/G1 phase, elevated proportions of cells in S phase and G2/M phase, as well as down-regulated expressions of JAK2, STAT3, p53, Bax, Caspase-3, pJAK2 and pSTAT3, elevated expression of Bcl-2 (all P<0.05). The most remarkable changes were observed in H/R+AG490 (50\u03bcM)+tanshinone IIA (50\u03bcM) group. Tanshinone IIA may attenuate H/R-induced MMEC apoptosis in rats by inhibiting the JAK2/STAT3 signaling pathway and regulating the expressions of p53, Bax, Caspase-3 and Bcl-2, which may provide a protective effect of tanshinone IIA for MMECs.
```

marcovzla:
This PR changes `countSemanticNegatives` so that it no longer tries to find the shortest path between every combination of tokens between a trigger and an argument. Instead we now search for the shortest path between the heads only.

Closes #426

### DIFF
--- a/main/src/main/resources/org/clulab/reach/biogrammar/events/neg-reg-only_template.yml
+++ b/main/src/main/resources/org/clulab/reach/biogrammar/events/neg-reg-only_template.yml
@@ -10,7 +10,7 @@
   example: "Cells rescued with wild type Gab1 exhibit Gab1 tyrosine phosphorylation in response to EGF inhibition"
   action: ${ actionFlow }
   pattern: |
-    @controlled:SimpleEvent (?<trigger> in response to (treatment with)?) @controller:${ controllerType } [lemma=/${ triggers }/ & tag=/^NN/]
+    @controlled:SimpleEvent in response to (treatment with)? @controller:${ controllerType } (?<trigger> [lemma=/${ triggers }/ & tag=/^NN/])
 
 
 # This applies only to non-nested regs
--- a/main/src/main/scala/org/clulab/reach/darpa/DarpaActions.scala
+++ b/main/src/main/scala/org/clulab/reach/darpa/DarpaActions.scala
@@ -460,6 +460,11 @@ object DarpaActions {
       triggerHead <- DependencyUtils.findHeadStrict(trigger.tokenInterval, trigger.sentenceObj)
       argHead <- DependencyUtils.findHeadStrict(arg.tokenInterval, arg.sentenceObj)
     } yield deps.shortestPath(triggerHead, argHead, ignoreDirection = true)
+
+    //println("Trigger: " + trigger.start + " -> " + trigger.end + " " + trigger.label)
+    //println("Argument: " + arg.start + " -> " + arg.end + " " + arg.label)
+    //println(s"Shortest path: ${shortestPath.get.mkString(", ")} in sentence ${trigger.sentenceObj.words.mkString(", ")}")
+
     shortestPath match {
       case None => Nil
       case Some(path) =>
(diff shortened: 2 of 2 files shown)


myedibleenso:
Looks like one failing test in `TestPaperReader`: 
>- should contain 1 negative regulation and 1 phosphorylation *** FAILED *** (98 milliseconds)

MihaiSurdeanu:
One test fails:

[info] Cells expressing ErbB3 show tyrosine phosphorylation in response to RAS inhibition
[info] - should contain 1 negative regulation and 1 phosphorylation *** FAILED *** (98 milliseconds)
[info]   List() had size 0 instead of expected size 1 (TestRegulationEvents.scala:405)

marcovzla:
It seems that there is a rule called `Negative_regulation_in_response_to` that matches the regulation in the test, but then the label is flipped using `countSemanticNegatives` which seems to be doing the right thing. So I would argue that previously we needed the `Negative_regulation_in_response_to` rule because `countSemanticNegatives` wasn't working as intended.

Here is a link to the rule: https://github.com/clulab/reach/blob/df3f84ea3dc60e8ad436a9ca405384acadf3d094/main/src/main/resources/org/clulab/reach/biogrammar/events/neg-reg-only_template.yml#L6

MihaiSurdeanu:
I'm looking into this now.

MihaiSurdeanu:
I think I figured it out: we are using "inhibition" twice: once in the rule Marco mentions, and once in countSemanticNegatives(), where we exclude words only if they are part of the trigger. The trigger for that rule is "in response to", which is wrong. I'll fix this.

MihaiSurdeanu:
@marcovzla: fixed and merged.

marcovzla:
:tada: Thanks!

Issue #516, opened 2017-05-26
Positive_activation_syntax_1_verb misclassifies a catalysis as a positive activation


johnbachman:
The sentence

```GAPDH catalyzes the conversion of glyceraldehyde-3-phosphate to 1,3-bisphosphoglycerate```

Leads to an extraction of a positive activation event by ```Positive_activation_syntax_1_verb```, with GAPDH as the controller and glyceraldehyde-3-phosphate as the controlled. However GAPDH is actually here catalyzing the reaction that *consumes* glyceraldehyde-3-phosphate and *produces* 1,3-bisphosphoglycerate.

myedibleenso:
closed by #517

PR #517, opened 2017-05-26
Remove catalysis from activation rule trigger

bgyori:
This PR, I hope, fixes issue #516 by removing catalysis from the trigger, which currently takes all posTriggers from events_master.yml.


bgyori:
This approach is only adequate if this rule applies only if the controlled is an entity, not another event. Is that the case here?

myedibleenso:
Nice, @bgyori!  In the future, this might become a variable that references a disjunction of several disallowed terms.

PR #533, opened 2017-07-27
Update for Akka-based processors server

hickst:


hickst:
Oh rats....this is dependent on Processors 6.0.7-SNAPSHOT. I assume we have to release a new Processors before this can be pulled.

PR #534, opened 2017-07-28
Update to use Akka-based processors server

hickst:


MihaiSurdeanu:
Can you please add a usage example to the readme?

PR #543, opened 2017-09-13
Trying to debug Travis-only failure

hickst:
Start by increasing memory (via `sbt`) to 7G to see if "Parsing of sentence failed, possibly because of out of memory." errors are real or not.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,3 +1,3 @@
--J-Xmx5G
+-J-Xmx7G
 -J-XX:+CMSClassUnloadingEnabled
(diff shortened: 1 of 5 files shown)


bgyori:
This might be helpful, it might be the same issue: https://github.com/sorgerlab/indra/issues/290

hickst:
Thanks Ben! I think the problem is the same, but your solution is (according to what I read) only applicable to the **container** run of Travis. We're using the VM run and our Travis build file abdicates all responsibility to `sbt`, so the solution that worked in my test was to increase the allocated memory in `.sbtopts`

PR #561, opened 2017-11-15
Test PR: changing Travis to run script

hickst:
If this works (and tests pass, of course) then we can actually merge this branch.


hickst:
The Travis script looks like it is working now but the PR is dependent upon releasing the next version of Processors and updating Reach main/build.sbt file for that release.

PR #577, opened 2018-02-11
Handle loops which are produced by BioCluProcessor

kwalcock:


MihaiSurdeanu:
1. What is the logic for line 195:
case loop if loop.head == i || heads.contains(loop.head) => highest
?

2. Why is line 24 in TestFindHeads commented out?

3. Can you please add a couple more tests in TestFindHeads?

(Longer term, this findHead* code should be moved to processors)

kwalcock:
1. if loop.head == i is for loops to self in order to avoid violating the invariant.  The original has valid.contains(i), which is less efficient.  heads.contains(loop.head) is for loops to something else already visited.  Does that make sense?

2. The test doesn't know which processor is running and different ones may get different results.  The important part is that it doesn't hang.  I could make a TimeLimitTest of it, but timing is a little tricky.

3. Sure.  I'll see if there are any more loops in the test suite (according to the BioCluProcessor) and might also use some toy sentence and hack the graph into various kinds of improbable loops.

PR #582, opened 2018-05-30
Added "turnover" as a semantic polarity keyword

MihaiSurdeanu:


MihaiSurdeanu:
From @johnbachman's talk :)

johnbachman:
Nice! One note: "turnover" indicates not only the polarity, but also specifically indicates a regulation of amount, rather than activity. Can that be easily captured? Might be worth coming up with a list of these words (e.g., "expression", "stability", "turnover", "degradation", etc.) that are markers of amount regulation--if these aren't already in REACH.

MihaiSurdeanu:
What do you mean? Can you please give me a simple example?

johnbachman:
Sure! So "X promotes Y" is ambiguous as to whether X is regulating Y's activity or its amount. This distinction is important because it affects how the interaction is modeled: "X activates Y" means that "X changes the state of of Y such that it participates in downstream functions." "X increases Y" means that the state of Y doesn't change, but its abundance does.

Some (most?) triggers are ambiguous: e.g. "X promotes Y", "X stimulates Y", "X inhibits Y", etc. could all refer either to a regulation of Y's activity or its amount. However, if a modifier is added, e.g. "X promotes Y stability", "X stimulates Y turnover", "X inhibits Y expression" all indicate that it is the amount, not the activity of Y that is being affected.

PR #586, opened 2018-08-20
Exclude target from copula rule

bgyori:
This PR avoids "target" as a trigger for the Positive_X_copula_1 rule. I found that "A is the target of B" is extracted with the wrong argument order i.e. A becomes the controller and B the controlled instead of the other way around.

Here is an example from the shell before the PR
```
EVENTS:   1

MENTION TEXT:  BRAF is a target of KRAS
LABELS:        List(Positive_activation, ActivationEvent, ComplexEvent, Event, PossibleController)
DISPLAY LABEL: Positive_activation
	RULE => Positive_activation_copula_1
	TYPE => CorefEventMention

	TRIGGER => is a target
	controller (List(Gene_or_gene_product, MacroMolecule, Equivalable, BioChemicalEntity, BioEntity, Entity, PossibleController)) => BRAF
	controlled (List(Gene_or_gene_product, MacroMolecule, Equivalable, BioChemicalEntity, BioEntity, Entity, PossibleController)) => KRAS
CONTEXT: NONE
```
and after the PR
```
EVENTS:   0
```

### DIFF
--- a/main/src/main/resources/org/clulab/reach/biogrammar/events/pos-reg_template.yml
+++ b/main/src/main/resources/org/clulab/reach/biogrammar/events/pos-reg_template.yml
@@ -205,7 +205,7 @@
   label: ${ label }
   action: ${ actionFlow }
   pattern: |
-    trigger = [lemma=/be/ & tag=/^V/] []? []? [lemma=/${ triggers }/ & tag=/^N/]
+    trigger = [lemma=/be/ & tag=/^V/] []? []? [lemma=/${ triggers }/ & tag=/^N/ & !word=/(?i)^(target)/]
     controller:${ controllerType } = nsubj
     controlled:${ controlledType } = /prep_of|prep_in$/
 


MihaiSurdeanu:
Maybe the right fix here is to not use "target" as a regulation trigger overall? @enoriega: can you please look at this?

enoriega:
@MihaiSurdeanu I confirm that the latest pull request that removed target from the triggers takes care of this example. @bgyori can you confirm if this is no longer happening for other sentences using the latest commit of the `master` branch?

MihaiSurdeanu:
@maxaalexeeva: can you please look at this conflict when you have some time?

MihaiSurdeanu:
I'd like to release 1.5.1 after this is merged.

maxaalexeeva:
@MihaiSurdeanu, The only actual change in this PR is avoiding the word _target_ in trigger (!word=/(?i)^(target)/; the conflict is copy-pasted below), but in master, 'target' is already taken out of the list of triggers, so I do not think avoiding it here makes sense. 

Also, the current output for that sentence in master is `Events: 0`, and that's the desired output if I understand the original post correctly. 

```
<<<<<<< HEAD
    trigger = [lemma=/be/ & tag=/^V/] []? []? [lemma=/${ triggers }/ & tag=/^N/ & !word=/(?i)^(target)/]
    controller:${ controllerType } = nsubj
    controlled:${ controlledType } = /nmod_of|nmod_in$/
=======
    trigger = [lemma=/be/ & tag=/^V/] []? []? [lemma=/${ triggers }/ & tag=/^N/]
    controller:${ controllerType } = ${agents}
    controlled:${ controlledType } = /${genitive_case_marker}|${preps_in$_singleton}/
>>>>>>> master
```

MihaiSurdeanu:
Ok. Closing this PR since it's now obsolete.

PR #588, opened 2018-08-29
Webapp

enoriega:
Ported the Eidos visualizer into REACH. The functionality is complete sans color-coding.


marcovzla:
We don't need to check it in.
This project was originally created from a giter8 template: https://github.com/playframework/play-scala-seed.g8
The `.g8` directory stores some templates for the `g8Scaffold` sbt command. But we haven't installed the `sbt-giter8-scaffold` sbt plugging anyway, so we know nobody is using it.
I think we should remove it.
See this if you're interested: http://www.foundweekends.org/giter8/scaffolding.html

enoriega:
Just removed that directory. @marcovzla @MihaiSurdeanu please take a look

MihaiSurdeanu:
Looks good to me. Please merge.

PR #592, opened 2018-08-31
Reload

enoriega:
Reenabled the hot-reloading of the rules on `ReachSystem` and the `ReachShell`


enoriega:
BTW, hot-reloading comes for free on the web app

marcovzla:
some tests are failing

marcovzla:
I tried to use this and it seems the rules are not being reloaded in the shell.
I suspect `RuleReader` and `LoadableAttributes` are not playing well with each other.

enoriega:
The reason is that the resources are been reloaded from the `target` directory substructure instead of from the sources. I need to figure out a way to change that behavior or explicitly ask the code to read a specific directory with the bio grammar.

@marcovzla Is adding a a setting in application.conf that points to the directory to the bio-grammar acceptable?

PR #598, opened 2018-10-16
Polarity

enoriega:
Changes to the rules that improved the polarity changes


enoriega:
This should fix it @marcovzla

MihaiSurdeanu:
@enoriega: why do we need a log4j.properties here? I thought we were **not** using log4j, no?

enoriega:
@MihaiSurdeanu `slf4j` uses `log4j` as a backend. I created the file to control the output formatting, but I can remove it and won't make a difference for the users.

MihaiSurdeanu:
slf4j is just an API that can use many different backends. At the time, we configured it to use logback. We should try to keep the same backend...

enoriega:
@MihaiSurdeanu Just removed all references to `log4j`. Is this good?

PR #618, opened 2018-12-13
Implemented the hyphen-role-flipping feature

enoriega:
I created this PR to see where the feature breaks the existing tests as reported by travis and keep everyone in the loop to comment about the implications of the change.

Very soon all the possible errors will be fixed and the merge can happen


enoriega:
@MihaiSurdeanu This is weird. If you expand the details of the tests, some of them fail, despite being cleared by travis. I'm working on fixing those issues

enoriega:
How about now? @MihaiSurdeanu

PR #623, opened 2019-03-05
Update processors to 7.5.0

kwalcock:


MihaiSurdeanu:
Thanks! Let's wait for the tests, then I will release.

kwalcock:
:-(  I'm hoping that one of the other changes that went into processors at the last minute is the problem, or even into processors 7.4.4.  I'll start checking.

kwalcock:
Strange.  It says the tests passed, but there is a lot of red in the output...

MihaiSurdeanu:
I see them still pending...
Let's wait a bit.

kwalcock:
It *looks* like the problem was before the DuoHashTrie PR #311.  It's still being localized.

MihaiSurdeanu:
Thanks!

kwalcock:
The problem is with the AmbivalenceHash.  One line needs to be reverted to make tests pass.  It will be processors 7.5.1 then.

MihaiSurdeanu:
Ok

PR #625, opened 2019-03-12
Polarity engine

enoriega:
Refactored the linguistic polarity engine. Provide an API to implement the deep learning polarity engine and derivatives.


enoriega:
@MihaiSurdeanu what do you think about this API for the polarity engine?

enoriega:
Thanks

PR #632, opened 2019-06-08
Polarity engine

zhengzhongliang:
1, Change fatdynet version to 0.2.2
2, Use deep learning model as the default polarity engine for reach. When loading REACH, a trained deep learning polarity classifier will be loaded automatically.


MihaiSurdeanu:
@enoriega: can you please check this PR first?

enoriega:
There are some unit tests that are failing. All of them are related to polarity detection. The new deep learning engine predicts the opposite from the expected polarity on those tests. 

We need to discuss whether the model is wrong and needs to be retrained until those tests pass again, disable those tests or configuring the tests such that they use the deterministic polarity engine (as always) and write a new suite that tests the deep learning engine. 

At glance I prefer the last choice, however we may want to analyze this on our next meeting

MihaiSurdeanu:
Ok. Let’s discuss.
In principle, the tests should be correct.


On June 8, 2019 at 12:21:40 PM, Enrique Noriega (notifications@github.com)
wrote:

At glance I prefer the last choice, however we may want to analyze this on
our next meeting

enoriega:
@zhengzhongliang Some TODOs before accepting the PR:

1.  Talk about the tests with @MihaiSurdeanu on our next ASDF meeting
2.  Parameterize the files on lines https://github.com/clulab/reach/blob/polarity_engine/main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala#L23-L24
 Please add them to the resources directory and use the application.conf file to specify the paths.
3. Same as above for this file path that is the constructor's parameter: https://github.com/clulab/reach/blob/polarity_engine/main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala#L19
4. Same here: https://github.com/clulab/reach/blob/polarity_engine/main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala#L78
5. Please remove the default value for the parameter of this method: https://github.com/clulab/reach/blob/polarity_engine/main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala#L78
6. Replace the `println` statements by `logger.info` or `logger.debug` calls
7. I think this file was used as "scratch", however, we need a way to train the model on demand. Could you clean it up a bit and write some comments for the way it is meant to be used? https://github.com/clulab/reach/blob/polarity_engine/main/src/main/scala/org/clulab/polarity/ml/TrainDeepLearningModel.scala
8. [Optional] Parameterize these values: https://github.com/clulab/reach/blob/polarity_engine/main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala#L32-L37

MihaiSurdeanu:
Thanks Enrique!

These comments are all valid! Some simple rules of thumb:
- Hyper params should be set through a config file.
- Do not use println(). Instead use logger.debug().
- Do not have absolute file paths in the code anywhere.

zhengzhongliang:
Thanks Enrique! I will work on these today.

Mihai Surdeanu <notifications@github.com>于2019年6月18日 周二上午12:49写道：

PR #638, opened 2019-09-19
Try out a function

kwalcock:


kwalcock:
This is bash's retribution for me thinking, "That's not a real programming language."

PR #641, opened 2019-10-15
Polarity engine expand

zhengzhongliang:
Last week I found that some samples in the easy subset are not expanded very reasonably. It turns out that we were using Stanford dependency tree. I switched to the universal dependency tree in the week (for both easy and hard set), and the expanded boundary are much more reasonable. The deep learning classifier achieves the best performance so far on the unit tests (but still fails in about 16 examples). The Hybrid model passes all tests.


zhengzhongliang:
Hi Professor,
I finished most of them and have a little question about some points.

*1,(Question) Please turn the manual eval that starts at line 14 in
TrainDeepLearningModel into actual unit tests:*
 There are to main methods in this block: the "EvalDeepLearningModel"
method and the "ManualCheckModel". The "EvalDeepLearningModel" method just
evaluates the trained model on the dev part of the spreadsheet and it is
converted to unit test now. You can use the command sbt "testOnly
org.clulab.reach.TestDeepLearningPolarityEngine" to test it. It should
succeed if the classifier gets >0.87 dev f1 score.

The "ManualCheckModel" is for my debugging purpose. Most of the sentences
are copied from the existing unit tests, with various masking schemes. Do I
still need to add them? I think I need to either remove the function, or
leave it there.

*2, (Finished) All the models you generated and saved (and data files) such
as the SavedLSTM* files should be under resources/ if we use them in Reach.
I can't tell where these are now. If not used in Reach, they should not be
checked in*
The final saved model has been moved to the resources folder and tested.
Other saved models are deleted. The spreadsheet for training the model is
also moved to the resources folder.

*3, (Finished) Similarly, I see other big files like w2vvoc.txt that are in
the top directory in Reach? These files should not be checked in.*
Actually these two files are not used in our model. So I just deleted them.

*4, (Question) Why are there test classes called Failed_Test...? You
mentioned that all tests pass now with the Hybrid method, no? And, in any
case, failed tests should be marked with @ignore
<https://github.com/ignore>, not with a change in the class name.*
Those failed tests are for debugging purposes when using only the deep
learning model. I just copied them from the existing tests. In hybrid mode
there are no failed tests. So I guess I just need to delete them.

Best regards,
Zhengzhong

main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala
```diff
@@ -0,0 +1,659 @@
...
+  override def isFitted: Boolean = _isFitted
+
+  /**
+    * Gets the predicted polarity for the provided argument
+    *
+    * @param events Ordered sequence of events to get their polarity from
+    * @return Predictions of Polarity subclasses
+    */
+  override def predict(events: Seq[BioEventMention]): Seq[Polarity] = {
+    var predictions = Seq[Polarity]()
```
MihaiSurdeanu: Please turn this into an ArrayBuffer[Polarity] as we discussed.

main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala
```diff
@@ -0,0 +1,659 @@
...
+    else {NeutralPolarity}
+  }
+
+  def maskEvent(lemmas_raw:Array[String], event:BioEventMention,  maskOption:String): Array[String] ={
+    // filter out the edge cases where event has no controller or controlled.
+    var lemmas = lemmas_raw.clone()
+    if (event.arguments.contains("controller") && event.arguments.contains("controlled")){
+      // filter out the edge cases where controller or controlled is vector.
+      //if (event.arguments("controller").isInstanceOf[mutable.ArraySeq[Mention]] && event.arguments("controlled").isInstanceOf[mutable.ArraySeq[Mention]]) {
+      if (event.arguments("controller").isInstanceOf[Seq[Mention]] && event.arguments("controlled").isInstanceOf[Seq[Mention]]) {
```
MihaiSurdeanu: Please replace all instances of the Strings "controller", "controlled", and "theme" with global variables declared in the class DeepLearningPolarityClassifier, e.g., val CONTROLLER = "controller", etc.

main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala
```diff
@@ -0,0 +1,659 @@
...
+          }
+        }
+      }
+    }
+    lemmas
+  }
+
+  def predictManual(event:String, rulePolarity:Int): Unit= {
+    val words  = event.split(" ")
+    val y_pred = runInstance(words, rulePolarity)
```
MihaiSurdeanu: This call should be synchronized.

MihaiSurdeanu: assigned to @zhengzhongliang

main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala
MihaiSurdeanu: @zhengzhongliang

MihaiSurdeanu:
Hi @zhengzhongliang,

Answers to your points:
- 1. You can leave ManualCheckModel as is.
- 2, 3, 4: agreed. Thanks!

Other things:
- I added some comments in place for the class DeepLearningPolarityClassifier. Please address them.
- The unit tests are failing now, because I suspect the deep learning model does not fit in the RAM allocated by Travis. @kwalcock: can you please switch this process to Jenkins as well? Thanks!

After these we are ready to merge.

kwalcock:
Will do.

kwalcock:
It is not at all critical, but it might be nice to correct the spelling of behavior throughout, perhaps in a different branch.  There's quite a bit of "Behaivor".

MihaiSurdeanu:
@kwalcock: +1. Thanks for catching it!
@zhengzhongliang: please correct it.

zhengzhongliang:
Thanks for pointing this! I will do today.

Mihai Surdeanu <notifications@github.com>于2019年10月17日 周四上午9:11写道：

main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala
zhengzhongliang: Done.

main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala
zhengzhongliang: Done.

main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala
zhengzhongliang: Done.

zhengzhongliang:
Hi,
I solved the issues above and pushed the changes to the branch. The model passes all tests.

MihaiSurdeanu:
Merged, and branch deleted.

@kwalcock: can you please merge the jenkins branch in master as well?

PR #644, opened 2019-11-12
Context zhengzhong

zhengzhongliang:
Implemented regulation classifier using a strong baseline. The regulation type is determined by the number of occurrences of the keywords in the whole sentence. When the numbers of occurrences are the same, measure the distance from the keyword to the trigger. The final result is that it fails 17/183 tests. 10 of them are because regulation failure, 7 of them are because of polarity failure (the 7 can be addressed soon by regenerating the tests because the current tests are generated from the original spreadsheet using the linguistic engine instead of the hybrid engine).


zhengzhongliang:
Hi Professor,
I did the following changes to the code:
1, The keyword is now saved in a resource file called "experimental_regulation_type_keywords.csv". It is loaded at the beginning and the keywords of experimental regulations are loaded. This allows flexible keywords. The path of the file can be modified in the application.conf file. 
2, The processing of experimental regulation type is now not hard coded. It depends on the keywords and the types in the resource file. 
3, I changed the test file name to "ExperimentalRegulationTests".
4, There is only one thing hard coded: after the type of regulation is determined, reach will assign a type, and this type is defined in the "/reach/main/src/main/scala/org/clulab/reach/mentions/Modifications.scala" file. Since the classes are predefined, the types defined in this file may be different from the types loaded from the resource file. So currently I specify that if the regulation type is in the resource file but not defined as a class in the "Modifications.scala", I assign it an "Unassigned" type.

zhengzhongliang:
Yes it is ready to merge now.

MihaiSurdeanu:
@johnbachman: we now have experiment info in regulations. Can you please give it a try?

This can be easily customized in this resource file:
https://github.com/clulab/reach/blob/master/main/src/main/resources/experimental_regulation_type_keywords.csv

@zhengzhongliang, @zupon: can you please explain where this info is stored in the output?
Thanks!

PR #654, opened 2020-02-26
Exclude promoter binding as complex formation event

### LINKED ISSUE #653: Exclude promoter binding as complex formation event
There are many sentences of the form "A binds the B _promoter_". Here, a Binding event between A and B (implying physical interaction between two proteins) should not be extracted since promoter binding is a form of (possible) gene expression regulation.

To achieve this, `binding1b` should be changed to not allow an incoming `compound` edge from "promoter" on the second theme. Further, to handle "A binds the _promoter of_ B", "promoter" should not be allowed to be an incoming `nmod_of` to the second theme in `binding_syntax_23_verb`. I couldn't figure out the correct syntax to add these constraints and am therefore asking for help. I pushed two tests that should pass on the `no_promoter` branch: https://github.com/clulab/reach/blob/no_promoter/main/src/test/scala/org/clulab/reach/TestBindingEvents.scala#L476-L486

bgyori:
Resolves #653.

### DIFF
--- a/main/src/main/resources/org/clulab/reach/biogrammar/events/bind_events.yml
+++ b/main/src/main/resources/org/clulab/reach/biogrammar/events/bind_events.yml
@@ -21,7 +21,7 @@ rules:
   pattern: |
     trigger = [lemma=/bind|associate|ligate/ & tag=/^V/]
     theme1:BioChemicalEntity+ = ${adjective_clause} /${conjunctions}/{,2} | ${agents} /${noun_modifiers}|${conjunctions}/{,2} | </${passive_agents}/{,2} ${agents}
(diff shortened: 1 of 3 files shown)


MihaiSurdeanu:
Thanks @bgyori !
Jenkins should be back up. Let's see if tests pass, just as a sanity check for both jenkins and this brach.

johnbachman:
Should these events be extracted as positive regulations of amount? That is how they are usually interpreted.

bgyori:
They could be, though promoter binding may not always imply regulation, and also, the polarity of regulation would be undetermined. For now, this PR just removes the incorrect complex formation event.

bgyori:
On another note, though the PR is approved and all checks passed, the Merge button is still blocked so the permission has to do with something else. My guess is that "write" permission means that I can push branches to the repo, but there is an additional policy for protected branches (like master) that doesn't apply to me (which I'm totally okay with).

MihaiSurdeanu:
How about now? I changed some team settings.

PR #656, opened 2020-03-06
Biores uniprot update

MihaiSurdeanu:


MihaiSurdeanu:
@bgyori: I am merging your biores_uniprot_update into proc800, so we have only one branch to merge in master when we're done.

PR #660, opened 2020-03-18
Use resources instead of files when indicated

kwalcock:
This applies only to the DeepLearningPolarityClassifier class!


kwalcock:
@bgyori, would you like to give this a try before I break master with it?

PR #661, opened 2020-03-23
Address resource access outside development environment.

kwalcock:


kwalcock:
There is still a `ReadPapers.papersDir = src/test/resources/inputs/nxml/` in resources/application.conf of the main subproject, but it looks like all uses are in Apps which are probably intended to be run during development time when the resources are available as files.  I did not reprogram them.

PR #665, opened 2020-04-08
Switched to processor 8.0.0; BioNLPProcessors is now included in Reach.

MihaiSurdeanu:


MihaiSurdeanu:
@kwalcock: this PR fails because Travis can't find processors-models (inherited from processors-main), which now lives in artifactory: 

[error] (processors/*:update) sbt.ResolveException: unresolved dependency: org.clulab#processors-models;0.0.4: not found
...

Can you please take a look? Does this mean we need to add the extra repository to reach as well, even though this dependency is inherited from processors?
Thanks!!

MihaiSurdeanu:
@bgyori: once this hurdle is solved, everything will be up to date in master.
After this, I'd like to move bioresources as a subproject in reach, so everything is contained in this project.

kwalcock:
sbt seems to know about the transitive dependency, but it doesn't know how to satisfy it.  If I add
```
resolvers += "Artifactory" at "http://artifactory.cs.arizona.edu:8081/artifactory/sbt-release"
```
to processors/build.sbt in the reach project, then it seems to work.  I wonder if there is a way to have processors broadcast that it is dependent on a particular non-standard repository.  I'll google that.

kwalcock:
It looks like some of our large dependencies could have been specified like this, just pointing to the jar file.  They hint, however, that the resolver information can be added to the metadata and that would be useful.

Explicit URL 
If your project requires a dependency that is not present in a repository, a direct URL to its jar can be specified as follows:
```
libraryDependencies += "slinky" % "slinky" % "2.1" from "https://slinky2.googlecode.com/svn/artifacts/2.1/slinky.jar"
```
The URL is only used as a fallback if the dependency cannot be found through the configured repositories. Also, the explicit URL is not included in published metadata (that is, the pom or ivy.xml).

MihaiSurdeanu:
Thanks @kwalcock!

For now, I added our artifactory repo to reach/processors/build.sbt, so we don't have to modify processors. Let's see if this works.

MihaiSurdeanu:
Success!
Thanks!

kwalcock:
I'm experimenting with a factuality client client which has a transitive dependency on the factuality model.  I'll report back if I can get it to work.

kwalcock:
If this information, nearly 8 years old, is still correct, then sbt does not have "transitive resolvers" and anyone using processors (with sbt) would need to add Artifactory to their resolvers: https://stackoverflow.com/questions/10727079/how-do-transitive-resolvers-work-with-sbt/61112404#61112404.  I was not aware of the limitation.  It may be that other tools can do this transitively and so it might be good for processors to "export" this information about processors-model dependency.  That does not seem to happen automatically.  Some extra code is required in build.sbt of the project publishing with the to-be transitive dependency:
```scala
resolvers += "Artifactory" at "http://artifactory.cs.arizona.edu:8081/artifactory/sbt-release"

pomIncludeRepository := { (repo: MavenRepository) =>
  repo.root.startsWith("http://artifactory")
}
```
To see if that works, I'd probably need to test with maven or gradle, assuming they do the transitive resolving.

MihaiSurdeanu:
Adding the extra resolver is not too big of a pain, so I consider this low priority.

PR #669, opened 2020-04-09
Enable assembly again

kwalcock:


MihaiSurdeanu:
Thanks!!
The Travis build failed. I think we need to completely switch to Jenkins. Do you mind doing that too?
Also, what does it mean to disable the merging strategy in assembly? I read somewhere that this means that the fat jar will be a collection of jars rather than a unified one, which means that finding resources in it will be slower. Is that true?

bgyori:
Thanks, this works for me! One thing I noticed is the jar file went from ~1GB to 2.75GB in size, is that expected?

MihaiSurdeanu:
Yes, there are some big deep learning models now in there.

kwalcock:
Travis has been disabled via the webhook at github.  No source code changes were required.  The configuration file remains for potential future use.  The badge in the readme file has been rerouted to jenkins.

That's not my understanding of the merge strategy.  Files with the same names, but different contents, were found in two jar files from library dependencies.  They were not the usual suspects like licenses that can be combined.  We have to tell assembly what to do in that case.  I'm restoring the comment with which I once explained it.

MihaiSurdeanu:
Got it. Thanks!

PR #670, opened 2020-04-10
Restore comment explaining merge

kwalcock:
...and also test that travis is disabled


kwalcock:
This is looking OK, assuming that the assembly is functional.

PR #672, opened 2020-04-11
Add .gitattributes

kwalcock:
Prevent the model file from being treated as a text file.  It has internal offsets which will be incorrect if line endings change.


MihaiSurdeanu:
Thanks @kwalcock !

PR #674, opened 2020-04-11
Master cleaning 20200409

zhengzhongliang:
I saved the word to index and character to index file to two txt files so that they are loaded at runtime.


MihaiSurdeanu:
Thanks @zhengzhongliang !
I'll merge this after unit tests complete.

MihaiSurdeanu:
@zhengzhongliang: I don't think this works. In particular, you can't treat the w2i and c2i files as Files. You have to open them as resources. @kwalcock: what is the preferred way to do this now? I believe you added some utilities for this. Thanks!

zhengzhongliang:
I am checking them. Sorry about that, I remembered I tested them and they
worked, but let me figure out ...

MihaiSurdeanu:
I still see this output when Reach starts, which suggests it does not work:

```
20:49:44.077 [run-main-0] INFO  o.c.p.m.DeepLearningPolarityClassifier - Making vocabulary for deep learning model ...
20:49:44.082 [run-main-0] INFO  o.c.p.m.DeepLearningPolarityClassifier - Loading data from spreadsheet ...
20:49:44.560 [run-main-0] INFO  o.c.p.m.DeepLearningPolarityClassifier - Num. all samples: 67882
20:49:44.565 [run-main-0] INFO  o.c.p.m.DeepLearningPolarityClassifier - Num. all labels: 67882
20:49:44.621 [run-main-0] INFO  o.c.p.m.DeepLearningPolarityClassifier - Loading data finished!
20:49:44.795 [run-main-0] INFO  o.c.p.m.DeepLearningPolarityClassifier - Vocabulary build finished! W2I size 19231,  C2I size 63
20:49:44.823 [run-main-0] INFO  o.c.p.m.DeepLearningPolarityClassifier - Loading saved model SavedLSTM_WideBound_u_tag ...
20:49:45.963 [run-main-0] INFO  o.c.p.m.DeepLearningPolarityClassifier - Loading model finished!
```

MihaiSurdeanu:
And, in any case, you should never use `Files.exists` to check for resources, and open them as files. This works locally in your own checkout, but it stops working when the project is packaged in a jar.

zhengzhongliang:
I see. Sorry, I am fixing this by loading from resource.

kwalcock:
It sounds like it has been figured out.  If some file is in the resources directory and headed towards the jar file upon publication, then it needs to be accessed as a resource.  It looks like there is an option for saving the file as well.  Sometimes if this just saved file is supposed to supersede the version in the resource, then we check for the file and read that before resorting to the resource.  There is code for that in fatdynet somewhere.  Search for things like "jar:" or "file:" or check for test cases.  I think that processors makes use of that option.

One thing that we may start running into is that resources of the same name in the top directory of any jar file can start to conflict with those from other jar files.  Assembly starts complaining and even if not assembled, I'm not sure offhand which resource will be used.  We may want to start putting them into namespaces.  So, lstmPolarity2i.txt might eventually become org/clulab/polarity/ml/lstmPolarityC2.txt to avoid conflicts.  It's not critical yet.  It applies to our generic configuration files as well, like reference.conf.  (IMO)

zhengzhongliang:
Hi,
I think this is fixed now. I changed two parts so that the w2i and c2i are
loaded as resources.

1, Instead of using Files.exists, I use:

if (Option(DeepLearningPolarityClassifier.getClass.getResource(w2iPath)).isDefined
& Option(DeepLearningPolarityClassifier.getClass.getResource(c2iPath)).isDefined){

2, The w2i and c2i are loaded as resources:

val bufferedSource = DeepLearningPolarityClassifier.sourceFromResource(filename)

3, I tested the script on the polarity related tests and nothing is
broken. And it does not show building vocabulary from spreadsheets
anymore:

21:58:14.324 [pool-4-thread-1-ScalaTest-running-TestTranslocationEvents]
INFO  o.c.p.m.DeepLearningPolarityClassifier - Loading saved model
SavedLSTM_WideBound_u_tag ...

21:58:15.434 [pool-4-thread-1-ScalaTest-running-TestTranslocationEvents]
INFO  o.c.p.m.DeepLearningPolarityClassifier - Loading model finished!

21:58:54.100 [pool-4-thread-1-ScalaTest-running-TestTranslocationEvents]
INFO  e.s.nlp.pipeline.StanfordCoreNLP - Adding annotator pos


...




I have pushed the code to the branch master_cleaning_20200409 and made
a pull request. Please let me know if this does not work.


Best regards,

Zhengzhong

MihaiSurdeanu:
Thanks Zhengzhong! Looks good.

PR #679, opened 2020-04-23
Integrating ChEBI, GO and new stopwords from bioresources

bgyori:
This PR builds on https://github.com/clulab/bioresources/pull/31, https://github.com/clulab/bioresources/pull/32 and https://github.com/clulab/bioresources/pull/33 to add ChEBI as a source of chemical names in addition to PubChem, GO synonyms and names for biological processes, and a new list of stopwords.

### DIFF
--- a/export/src/test/scala/org/clulab/reach/export/TestApi.scala
+++ b/export/src/test/scala/org/clulab/reach/export/TestApi.scala
@@ -119,9 +119,9 @@ class TestApi extends FlatSpec with Matchers {
     hasEntity("ASPP2", results) should be (true)
   }
 
-  it should "return 4 positive activation and 3 phosphorylation results from NXML test" in {
+  it should "return 9 positive activation and 3 phosphorylation results from NXML test" in {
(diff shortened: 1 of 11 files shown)


MihaiSurdeanu:
@bgyori : this does not build. Is it because it depends on the new bioresources that hasn't been released?

bgyori:
Yes, it depends on 1.1.32-SNAPSHOT.

bgyori:
I am working on some other updates so it probably makes sense to wait before releasing bioresources and merging this.

bgyori:
There are actually more test failures that I can see now, the basic issue is that ChEBI has grounded entries for many very generic terms like `protein`, `inhibitor`, `solution`, etc. that aren't things we want to individually pull out as named entities, especially when they appear in a context such as `the Ras protein`. I am removing some entries from the ChEBI table in bioresources to improve on this issue.

bgyori:
Just changed to the released version of bioresources here so it should build now, let's see if all the tests pass.

PR #683, opened 2020-04-28
Conversion

bgyori:
This is a work in progress PR, originally by @johnbachman to add extraction rules for Conversion events. I rebased the old branch on master, updated the syntax of the rules to match the current syntax, and made some modifications to make test sentences work. There are several questions related to this:
- This commit: https://github.com/clulab/reach/commit/dcc3cbd9f4a6d0b25c5fc3dede2d95d23ad16a97 appears to be necessary to have any of the Conversion extractions work (this change was recommended by @danebell years ago as a workaround) but I am not sure anymore why it is needed and how to replace it with something that isn't a workaround (given that keeping this commit seems to break a number of other behaviors in Reach).
- Some of the rules still need tweaking because they overlap with existing ones, e.g., convert_with_controller  interferes with other phosphorylation extraction rules.
- Some of the rules could be reimplemented using variables


danebell:
I don't remember why that change bypassing the coreference module was needed, but making that change permanent seems suboptimal, not least because it still runs coreference without using the results. I'll see if I can figure out a more correct fix. Thanks for the ping :-)

bgyori:
One thing I don't understand is, what is it about Conversions that requires this change? Is it the fact that they have up to 3 arguments (controller, substrate, product)?

MihaiSurdeanu:
Thanks @bgyori and @danebell !
Yes, I don't understand either the need for that particular commit...
Also, I see that you made the 3 arguments mandatory. Is that because a conversion extraction without any of the 3 is not useful? This would certainly reduce the number of extractions.

MihaiSurdeanu:
@kwalcock : I don't understand why jenkins fails on this PR. I don't see any error messages in the log. When you have a moment, can you please take a look? Thanks!!

bgyori:
So what I see is that if we go by `resolved` instead of `grounded` in that line then the Event gets filtered out (not sure yet at what point exactly or why but my guess is it has something to do with the non-standard argument structure/naming). In terms of making the arguments mandatory: these events are really mostly useful if all the arguments are there, plus this also makes the rules more reliable / pick up less spurious stuff. We could relax this later if necessary.

kwalcock:
Looking now.

kwalcock:
It seems like Ben must be seeing errors and so do I.  Sometimes the errors are further up the buffer than is shown by default so that there's a "Full Log" link, but there seem to be plenty below that.  Maybe something, a browser cache, needed to be refreshed.

http://jenkins.cs.arizona.edu:8090/job/reach/job/conversion/2/consoleFull

MihaiSurdeanu:
Thanks @kwalcock!

Ben: it seems many tests are broken in this PR...


On April 28, 2020 at 9:31:00 PM, Keith Alcock (notifications@github.com)
wrote:

It seems like Ben must be seeing errors and so do I. Sometimes the errors
are further up the buffer than is shown by default so that there's a "Full
Log" link, but there seem to be plenty below that. Maybe something, a
browser cache, needed to be refreshed.

http://jenkins.cs.arizona.edu:8090/job/reach/job/conversion/2/consoleFull

—
You are receiving this because you commented.
<https://github.com/clulab/reach/pull/683#issuecomment-620982883>, or
unsubscribe

bgyori:
Yes, I'm aware of the test failures - they are due to that workaround with the events being filtered out that I'm asking for help finding an alternative to.

bgyori:
I figured out all the issues and this is now working. Probably could be tweaked further but I think this is a good start.

bgyori:
Yes, unless you see any issues or want me to change anything else!

PR #686, opened 2020-05-07
Integrate Disease entity type and grounding

bgyori:
This PR adds a new entity type called Disease, and integrates MeSH groundings for diseases from bioresources (https://github.com/clulab/bioresources/pull/34).


MihaiSurdeanu:
I just releases bioresources 1.1.33 with MeSH entities. Should be up on Maven in a couple of hours.

PR #687, opened 2020-05-18
Master crash fix 20200512

zhengzhongliang:
Fixed the dynet memory issue. 
 - The default memory allocated to Dynet is now 2048 Mb (so that each dynet block gets 512 Mb).
 - The total value of dynet memory can be set in the application.conf file, and each block will be assigned equal values of memory. 
 - The maximum dynet memory of each block is clipped to 768 Mb for now. 
 - I tested it by processing some papers. I tested 30 papers and no dynet memory issue is reported.

### DIFF
--- a/main/src/main/resources/application.conf
+++ b/main/src/main/resources/application.conf
@@ -56,7 +56,7 @@ polarity {
   savedModel = SavedLSTM_WideBound_u // SavedLSTM // SavedLSTM_WideBound
   w2i = /lstmPolarityW2i.txt // save the word to index dictionary of the lstm polarity classifier.
   c2i = /lstmPolarityC2i.txt // save the character to index dictionoary of the lstm polarity classifier.
-  dynetMem =  2048 // number should be in MB
+  dynetMem =  4096 // number should be in MB
 
 
   spreadsheetPath = /SentencesInfo_all_label_final_ExactRecur_ExpandBound_universal.txt
--- a/main/src/main/resources/application.conf
+++ b/main/src/main/resources/application.conf
@@ -20,7 +20,7 @@ outDir = ${rootDir}/output
 
 # this is the directory that stores the raw nxml, .csv, and/or .tsv files
 # this directory *must* exist
-papersDir = ${rootDir}/papers
+papersDir = ${rootDir}/papers/corpus_20200513_control_1  # change this directory for debugging
 
 
 # the encoding of input and output files
--- a/main/src/main/resources/application.conf
+++ b/main/src/main/resources/application.conf
@@ -56,7 +56,7 @@ polarity {
   savedModel = SavedLSTM_WideBound_u // SavedLSTM // SavedLSTM_WideBound
   w2i = /lstmPolarityW2i.txt // save the word to index dictionary of the lstm polarity classifier.
   c2i = /lstmPolarityC2i.txt // save the character to index dictionoary of the lstm polarity classifier.
-  dynetMem =  4096 // number should be in MB
+  dynetMem =  512 // number should be in MB
 
 
   spreadsheetPath = /SentencesInfo_all_label_final_ExactRecur_ExpandBound_universal.txt
--- a/main/src/main/resources/application.conf
+++ b/main/src/main/resources/application.conf
@@ -20,7 +20,7 @@ outDir = ${rootDir}/output
 
 # this is the directory that stores the raw nxml, .csv, and/or .tsv files
 # this directory *must* exist
-papersDir = ${rootDir}/papers/corpus_20200513_control_1  # change this directory for debugging
+papersDir = ${rootDir}/papers/corpus_20200513  # change this directory for debugging
 
 
 # the encoding of input and output files
--- a/main/src/main/resources/application.conf
+++ b/main/src/main/resources/application.conf
@@ -97,7 +97,7 @@ logging {
 # restart configuration
 restart {
   # restart allows batch jobs to skip over input files already successfully processed
-  useRestart = true
+  useRestart = false //TODO 20200525: changed this for debugging.
   # restart log is one filename per line list of input files already successfully processed
   logfile = ${outDir}/restart.log
 }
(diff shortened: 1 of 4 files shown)


main/src/main/resources/application.conf
```diff
@@ -20,7 +20,7 @@ outDir = ${rootDir}/output
 
 # this is the directory that stores the raw nxml, .csv, and/or .tsv files
 # this directory *must* exist
-papersDir = ${rootDir}/papers
+papersDir = ${rootDir}/papers/corpus_20200513_control_1  # change this directory for debugging
```
bgyori, line 23: You might want to change this back to the default before merging

MihaiSurdeanu:
@bgyori: once @zhengzhongliang fixes these issues, would you mind rerunning your docs?
Note that we no longer see dynet issues, but there are NxmlReader exceptions on a few docs. This may be caused by incorrect NXML formatting in these docs, or some issues in our Nxml reader, which we should fix.

MihaiSurdeanu:
@bgyori: just to be sure: are you running the branch or master? This has not been merged yet.

main/src/main/resources/application.conf
zhengzhongliang: Oh sorry, I will change this to the default after I fix this.

bgyori:
Yes, I'm on this branch and tried to double and then quadruple the memory limits. Currently, I'm testing with 
```
dynetMem =  8192
```
in application.conf and
```
var dynetBlockMem = "2048"
```
in DeepLearningPolarityClassifier.scala. I also have application.conf pointing to an input folder with the 224 txt/nxml documents I sent earlier and I'm still getting the same error. I have been doing this in a Dockerized dev environment but I can try outside it too to see if that makes a difference.

MihaiSurdeanu:
@zhengzhongliang: please let us know if you can replicate the bug in your run.

bgyori:
Tried on my Mac outside Docker and I'm getting the same crash. @zhengzhongliang in your comment you mentioned having tried this on 30 test documents but is the reading working for you with the full corpus I sent?

zhengzhongliang:
> Tried on my Mac outside Docker and I'm getting the same crash. @zhengzhongliang in your comment you mentioned having tried this on 30 test documents but is the reading working for you with the full corpus I sent?

Not yet. I am testing it on the full corpus now.

MihaiSurdeanu:
@zhengzhongliang: were you able to replicate this crash? Thanks!

zhengzhongliang:
Yes I am able to reproduce that. Here are some experiments I have done and
the results. It looks like this happens pretty randomly, and I am looking
at possible reasons.

MihaiSurdeanu:
Thanks! I am trying to understand if this is a memory issue, a parallelism issue, or something else. Can you please repeat the jenny experiment (say the one with 512MB) multiple times to see if it crashes on the same paper? Also, what happens if you repeat this experiment with different random seeds?

zhengzhongliang:
Ok, I will do it now.

zhengzhongliang:
Hi @bgyori ,
We fixed a synchronization issue in the code. Now the program can process all papers without the dynet error (134 succeeded, the others failed due to the format issue). In the next few days I will test the code with other memory settings and on other machines for more times to make sure it is robust, but you can also try to test it on your machine now.

Best regards,
Zhengzhong

bgyori:
I just finished the test run and it worked, thanks!

MihaiSurdeanu:
@zhengzhongliang: are you ready for the PR? Note that some unit tests seem to fail in jenkins.

zhengzhongliang:
I am almost done. I am running some additional tests on river to make sure
it can succeed in other memory settings. I will comment in the thread when
it is ready to merge (should be tonight).

zhengzhongliang:
Hi @MihaiSurdeanu , 
I ran it on river with different dynet memory, and they all succeeded. This is ready to merge now.

Best regards,
Zhengzhong

main/src/main/scala/org/clulab/polarity/ml/DeepLearningPolarityClassifier.scala
```diff
@@ -61,11 +62,16 @@ class DeepLearningPolarityClassifier() extends PolarityClassifier{
     HIDDEN_SIZE = config.getInt(configPath+".HIDDEN_SIZE")
     MLP_HIDDEN_SIZE = config.getInt(configPath+".MLP_HIDDEN_SIZE")
     N_EPOCH = config.getInt(configPath+".N_EPOCH")
+
+    dynetBlockMem = scala.math.min(config.getInt(configPath+".dynetMem")/4, 768).toString // set dynet block memory, max block memory is set to 768 for now.
```
MihaiSurdeanu, line 66: This still doesn't seem to be right. dynetMem is set to 512MB in the config file. Either increase it to 2048, or remove the division by 4 here.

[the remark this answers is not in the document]
zhengzhongliang: Hi Professor,
I think the current one is correct. dynetMem exists in the config file, it is the total amount of memory that should be used by Dynet (i.e., by all the four blocks of dynet). dynetBlockMem is dynetMem/4, and dynetBlockMem is directly used to initialize the four memory blocks of Dynet. In running, I can clearly see that when I change the dynetMem in config file, the memory allocated to dynet will change. So I think the current thing works.

zhengzhongliang: I want to double check, the dynetMem is set to 512 Mb, and each dynet memory block gets 128 Mb. But since there are 4 blocks, the total amount of memory allocated to dynet is still 512 Mb, which is the previous default setting. May I know if I need to increase the total dynet memory to 2048 Mb (which is 4 times of the previous total dynet memory)? 

Code to initialize dynet memory: 

initializeDyNet(mem = dynetBlockMem+","+dynetBlockMem+","+dynetBlockMem+","+dynetBlockMem)

Terminal output when running:  

[dynet] allocating memory: 128,128,128,128MB
[dynet] memory allocation done.

MihaiSurdeanu: Yes, that's what I'm saying. You should increase it to 2048 Mb in the config.

zhengzhongliang: Changed to 2048 Mb.

MihaiSurdeanu:
Thanks @zhengzhongliang !

zhengzhongliang:
Sure. Thanks for you patience!

PR #688, opened 2020-05-21
Added missing ai.lum.common dependency

myedibleenso:
`ai.lum.common` is used in several places in a couple of subprojects, but it isn't included as an explicit dependency anywhere.

This can lead to compatibility issues in other projects that depend on reach, but use different versions of `ai.lum.common`.


myedibleenso:
Closing in favor of #691 .

PR #689, opened 2020-06-24
fix synchronization issue

zhengzhongliang:
Hi Professor,
I changed all the synchronization code in the LSTM polarity classifier to the style you suggested in the processor code. I did this by defining an empty "DyNetSync" object in the file (you should be able to see in the commit comparison). The new code passed all the polarity tests. Please let me know if there are anything to change.

Best regards,
Zhengzhong


MihaiSurdeanu:
Thanks! This better and it's Ok for now.
But, once we release the new processors, you should synchronize with the DyNetSync object *in processors* rather than create a new one here. This way, you are protected against other programs who may use DyNet at the same time.
I will let you know when we release processors.

@kwalcock: just FYI.

zhengzhongliang:
Got it. Thanks!

PR #693, opened 2020-09-03
Proonto

JakeWolfe:
Protein ontology protein fragment entity integration.


MihaiSurdeanu:
@JakeWolfe: this PR fails because it depends on the unreleased bioresources version. Once that one is merged and released (I will ping you when it is), please update the version number for bioresources, and make sure the PR passes the tests.

@bgyori: this is the PR that uses the new bioresources with protein fragments. Can you please take a look at the unit tests to see if they make sense to you? 

Thank you both!

MihaiSurdeanu:
This will continue to fail until we release bioresources. @bgyori : should we do it now, or wait until we have the Uniprot fragments in there?

kwalcock:
I'm rerouting this for a merge into a development branch, because quite a few separate changes need to be collected.

PR #697, opened 2020-10-14
Update sbt and plugins

kwalcock:


kwalcock:
This will be checked again with the merge from development to master.

PR #703, opened 2020-10-16
Bump bioresources to v1.1.34

kwalcock:


kwalcock:
It's still failing with OOM.  Will try a different strategy.

PR #707, opened 2020-10-16
Update Scala in build.sbt and docker

kwalcock:


kwalcock:
The problem is with the API test.  Two tests were running at the same time so that the port was already in use.  This will be tested again in the development branch anyway.

PR #728, opened 2021-01-19
Remove references to separate HGNC resource

bgyori:
This corresponds to https://github.com/clulab/bioresources/pull/56 and should be tested/merged once https://github.com/clulab/bioresources/pull/56 is merged and released. It removes references to a separate HGNC resource file.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 #Changes
 =======
++ **1.6.2** - Update bioresources to 1.1.38
 + **1.6.2** - Update bioresources to 1.1.36 and processors to 8.2.4.
 + **1.6.2** - Fix bugs related to OS and library dependency compatibility, bounds checking, serialization of newer Modifications, and output consistency
 + **1.6.2** - Added the `assembly` subproject back. The `arizona` and `cmu` formats are supported again.
--- a/main/src/main/scala/org/clulab/reach/grounding/ReachEntityLookup.scala
+++ b/main/src/main/scala/org/clulab/reach/grounding/ReachEntityLookup.scala
@@ -120,7 +120,6 @@ class ReachEntityLookup {
   val proteinSeq: KBSearchSequence = extraKBs ++ Seq(
     StaticProteinFragment, // TODO: for now fragments have higher priority
     StaticProtein,                          // Uniprot proteins
-    StaticGene,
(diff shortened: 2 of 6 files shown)


MihaiSurdeanu:
@kwalcock : this should be merged once the latest PR in bioresources is released. Thanks!

Also, I am discussing with @enoriega to merge bioresources as a subproject into Reach, so we don't go through this dance everytime.

kwalcock:
Yes, as soon as the previous PR passes its checks.

kwalcock:
@bgyori, if I merge master into this and then run the tests with the locally published hgnc_up branch of bioresources (straight or merged with its master branch), there is one test failure.  Perhaps there is an interaction with the previous PR that was merged in the meantime.

```
[info] - should match expected grounding IDs in the text p53, RAC1, RhoA, ROCK1, and SAF-1 GGPs. *** FAILED *** (11 milliseconds)
[info]   false was not true (TestOverrides.scala:188)
```

bgyori:
Yes, I noticed that test failure locally but when I pasted the sentence into the shell, I got all the right groundings so I wasn't sure what to make of it. I was having trouble debugging in IntelliJ (send a separate email about that) so couldn't figure out what the specific issue was. Do you have a sense of which test condition is failing?

kwalcock:
Now I have the same IntelliJ problem.  There was no sbt tab on the right side of the screen, which means IntelliJ doesn't think it's an sbt-built program and probably ignores dependencies.  I think that it has to do with the .bsp directory which newer versions of sbt seem to be using.  I erased that and also .idea and the tab reappeared.

kwalcock:
It looks like grounding P07998 is failing.

bgyori:
I see, I can reproduce that now (and IntelliJ is working again after "Reimport All sbt Projects"), it looks like there is an ambiguity for `RAC1` 
```
RAC1»···P07998»·Homo sapiens
RAC1»···P63000»·Homo sapiens
```
and for some reason the wrong one is chosen now. I'll have to add an override in bioresources for this one.

kwalcock:
I'm glad you know what to do.  I'll get the hang of it eventually.

bgyori:
Okay, I pushed the change to bioresources and the test passes for me now.

Issue #730, opened 2021-01-20
Merge bioresources into REACH


enoriega:
Every time we make an update to the [Bioresources](https://github.com/clulab/bioresources/) project, we need to bump the version number and publish it to maven central before being able to integrate the change into reach.

Given that reach is the only _consumer_ of bioresources, it makes sense to merge the repositories to streamline the process.

This will be done by migrating bioresources' sbt project as a subproject in reach's, similar export or  assembly

MihaiSurdeanu:
+1!

bgyori:
Do you think there's a way to preserve the git history of the bioresources repo while migrating? I pretty often look at the history / blame there to figure out what happened when and why so it might be useful. (Maybe something like this: https://medium.com/@leyanlo/how-to-move-one-git-repository-into-a-subdirectory-of-another-with-rebase-2b297b628c57, or making use of https://git-scm.com/book/en/v2/Git-Tools-Submodules).

enoriega:
I'll take a look. Worst case, we can trace back the changes in the history of the stand alone repo. We need to keep it to support older versions

enoriega:
I am a little wary of the solution described in https://medium.com/@leyanlo/how-to-move-one-git-repository-into-a-subdirectory-of-another-with-rebase-2b297b628c57 because it involves overwriting git's history, I think it is too risky, but did make a test with the submodules and it works relatively well.

I created branch `bioresources_submodule` that has the following changes:

- Added the bioresources repo as a submodule of REACH
- Modified the `build.sbt` files to import bioresources as a subproject instead of a dependency

Doing this has the least friction in terms of changes, because we get to preserve the same repositories (and each's history too). Bioresources changes would still be done in its own repo. The downside is of a bit of overhead when using `git` at the moment of the initial clone or when updating bioresources.

After cloning the repo, the user would have to execute:

```
git submodule init
git submodule update --progress
```

And to import changes from bioresources, pull with an additional flag:

```
git pull --recurse-submodules
```

All of these represents a little extra effort by the end user and may be a source of confusion. The alternative would be to copy the files into the main repository, which will result in a seamlessly transition at the cost of losing the history of bioresources.

@MihaiSurdeanu @kwalcock @bgyori  what do you think?

MihaiSurdeanu:
I agree that it's not worth porting the commit history from bioresources, since bioresources will continue to exist as a repo, and all those git commands scare me. But if @kwalcock thinks differently, I defer to him.

[...]

PR #734, opened 2021-01-21
Bioresources

enoriega:
Integrated `bioresources` into reach. This PR addresses issue #730. It is a copy of the files, reconfiguration of `build.sbt` and update of the shell scripts to account for the new directory structure.

The commit history is not carried over, but will remain available in the legacy [`bioresources`](https://github.com/clulab/bioresources/) repository


MihaiSurdeanu (approved):
LGTM, with a couple of changes.
I propose to NOT copy bioresources/CHANGES.md over. Leave it in its original repo.
But update CHANGES.md for reach with one line that indicates which bioresources version was copied over.

kwalcock:
Being less involved in bioresources' past, I don't have a strong opinion about preservation of the history, but perhaps it would be useful to review the repercussions for others.  Version control systems in general and git in particular don't seem to do a good job tracking changes across renaming and moving of files (a pet peeve of mine).  What will this look like?

MihaiSurdeanu:
I don't know... But git scares me, and I would prefer to keep it simple. I think it's up to us to document changes in KB file names when they occur. Which means that somebody who wants to track git changes to a file, may need to do an extra step to map the new file to the old history in the original bioresources repo.

bgyori:
Given that the original files will remain in the Bioresources repo, we can look at the history up to this PR there. We will also be able to track new history in this repo after the move. The only issue is that we can't do e.g.,, git blame across both the "new" and "old" history of files - not a huge issue I think. Further, many of the files are stored as `.tar.gz`, so it's hard to track exact content changes using git tools anyway. So I think moving over the files like this while keeping the original repo available is probably okay.

kwalcock:
It looks like the relevant bioresources people have weighed in.  It's easy for me to press the button.  I'll count to 100 first, though, just in case.

MihaiSurdeanu:
98, 99, ...

kwalcock:
...in binary.  It takes longer.

kwalcock:
On 109 I remembered that bioresources/CHANGES.md needed to be removed and the main CHANGES.md updated.  That will be separate now.

enoriega:
The remaining tasks are taken care of in PR #735

PR #735, opened 2021-01-22
Removed bioresources' CHANGES.md and updated reach's CHANGES.md

enoriega:
The lingering changes requested on #730


kwalcock:
BTW I'll get the testing sorted out.  Recently there were three tests being conducted and now only one.  The settings are confusing.  I'm aiming for two: the branch as is and as it would be after being merged with master.

PR #739, opened 2021-02-09
Context Module

enoriega:
This is the first PR related to adding the SVM context engine as a sub-module to REACH.

It has the following elements:

- Created a context submodule for the SVM implementation of the context engine. This includes all the auxiliary code to extract features for ML, parse the dataset files and training scripts. The original context engine trait and the deterministic context engine still live in the main module to avoid a circular dependency problem.
- Added the feature extraction code to the context sub-module
- Added the dataset parsers to the context sub-module
- Added a skeleton implementation to the SVM context engine to the context sub-module

Still TODO:

- Implement the training script using processors' SVM wrapper
- Implement the SVM context engine
- Write unit tests

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -96,10 +96,16 @@ lazy val processors = project
 lazy val main = project
   .settings(commonSettings:_*)
   .dependsOn(processors % "test->test;compile->compile")
+//  .dependsOn(context  % "test->test;compile->compile")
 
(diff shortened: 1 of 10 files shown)


MihaiSurdeanu (approved):
Looks clean! Please merge.
Thanks @enoriega !

PR #744, opened 2021-04-06
Uncompressed bioresources kb files

enoriega:
Expanded the gz files and modified the code to work this way. After this change, editing the KBs should be easier

PR #745, opened 2021-04-06
Fixed the tests

enoriega:
This commit fixes the errors in the tests from PR #744


bgyori:
One more thing I noticed is that since the UniProt file is now spit in 3 parts, I'll have to change the update script to dump into the appropriate file(s), so I'll look into that later.

MihaiSurdeanu:
Good point.


On April 6, 2021 at 12:48:04 PM, Benjamin M. Gyori ***@***.***)
wrote:

One more thing I noticed is that since the UniProt file is now split in 3
parts, I'll have to change the update script to dump into the appropriate
file(s), so I'll look into that later.

—
You are receiving this because you are subscribed to this thread.
<https://github.com/clulab/reach/pull/745#issuecomment-814394370>, or
unsubscribe

enoriega:
Correct. I had to do this in order to get around the single file size limit of 200MB

MihaiSurdeanu (approved):

PR #747, opened 2021-04-06
Remove duplicates from overrides

### LINKED ISSUE #746: Remove duplicate entries from Bioresources
I am about to remove the duplicate entry for E3:

```
['E3', '5756', '', 'pubchem', 'Simple_chemical'],
['E3', 'E3_Ub_ligase', '', 'fplx', 'Family']]
```
@MihaiSurdeanu  Which one is the correct entry?
Also, I detected more entities with duplicate but identical entries in `NER-Grounding-Override.tsv`. I will remove the redundant ones.

Now that I am on this, should I look for duplicate entries in the other KB files?

bgyori:
Fixes #746. Except for `E3` where there is a genuine ambiguity that had to be resolved (I chose the protein family sense there), all the other duplicates were just redundant rows that could be deleted.


enoriega:
The tests that fail here are fixed by PR #745. Perhaps we should merge that branch into master and then here to re-run Jenkins @kwalcock @MihaiSurdeanu

enoriega:
I edited the tests to account for the new grounding of E3 (`Simple_chemical` -> `Family`). We can go back to the way they were if @MihaiSurdeanu or @bgyori think so necessary

PR #748, opened 2021-04-07
Generate KBs as SBT Task

enoriega:
Created an SBT task (generateKBs) to update the NER files from the KB files. 
This task supersedes the shell scripts that had to be run outside of SBT for the same purpose.

Tun run it, just call `sbt generateKBs`


bgyori:
You might want to also update the instructions here: https://github.com/clulab/reach/blob/master/bioresources/README.md#updating-the-ner-files

enoriega:
@MihaiSurdeanu @kwalcock any comments before merging?

kwalcock:
Thanks for the reminder.  Please stand by.  I think it's absolutely great that this works through sbt rather than a shell command.  I'll try it out momentarily.

kwalcock:
Very handy.  I checked Linux and Windows.  In general I'd like to reduce the reliance on shell scripts because they aren't portable.  runAllTests.sh might be up soon.  I often just run `sbt test` and forget that not everything or enough gets tested.

enoriega:
That's certainly a good candidate. Each test is a native `sbt` task, so it should be possible compose them into a single task.

PR #751, opened 2021-05-05
Enable pre-annotation of document objects

enoriega:
I wrote this feature to support my own use case, but might as well be useful for anyone with large-scale annotation tasks.

To avoid generating Document objects every time the grammar changes, I added a class, `AnnotationsCLI` that mimicks `ReachCLI`, but stops short of running the grammar and serializes the Document and FriesEntry objects.

ReachCLI was enhanced to support "reading" ser files and pick it up from there, running the grammar over the pre-annotated documents and extracting the mentions.

If this makes sense to integrate into the master branch, I will update the documentation and write unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,10 @@ The most common use of Reach is to process a directory containing one or more pa
 
 Some configuration is necessary before running Reach. Please refer to the [Running Reach](https://github.com/clulab/reach/wiki/Running-Reach) Wiki page for detailed information on configuring and running Reach.
 
+## Pre-processing a directory of `.nxml` papers
+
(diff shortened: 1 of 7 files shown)


MihaiSurdeanu:
I agree this is very useful!
But there is a lot of code in here that is redundant with ReachCLI... Can you re-organize this code so there is no duplicated code? Maybe have a base CLI class that is extended by both ReachCLI and AnnotationsCLI?

enoriega:
@MihaiSureanu Done. I can do more effort and refactor similarly the Run*CLI classes, but don't think it is worth to. I will go ahead and write doc + tests in the coming days

MihaiSurdeanu (approved):
Looks great! Thanks!

PR #752, opened 2021-05-05
Move most of decoupling logic to processors

kwalcock:


kwalcock:
These changes accompany https://github.com/clulab/processors/pull/477.  That is, if we are allowed to change processors in order to support this new reach functionality, this is how I'd do it just now.

Processors has been updated to 8.3.3-SNAPSHOT here, so this can't be used directly until the processors change is merged and processors is republished.

The second commit includes a manual reordering of the label priorities.  This needs to be fixed with a design or convention change of some kind.  I'll defer at least temporarily to others on that.

kwalcock:
@enoriega, so you know among other things how that map was turned into a sequence.

processors/src/main/scala/org/clulab/processors/bionlp/ner/KBGenerator.scala
```diff
@@ -98,59 +95,16 @@ object KBGenerator {
...
-      .filter(_.nonEmpty)
-      .sorted
-      .distinct
+  def processKBFiles(caseInsensitiveMatching: Boolean): Seq[StandardKbSource] = {
+    val kbEntries = loadFromConf()
+    // This should only be performed on small collections.
+    // It is important to keep both the labels and the KbEntries in the same order.
+    val actualLabels = kbEntries.map(_.neLabel).distinct
+    val expectedLabels = Seq("Gene_or_gene_product", "Family", "Cellular_component", "Simple_chemical", "Site", "BioProcess", "Disease", "Species", "CellLine", "TissueType", "CellType", "Organ")
+    val kbEntriesSeq = expectedLabels.map { label => kbEntries.filter(_.neLabel == label) }
```
enoriega, line 104: This line and line 103 should go away for two reasons:
1. `loadFromConf` will guarantee the entries come in the order specified by the user in the configuration
2. A new KB file with a previously unseen label can be added by a user

I am sorry for leading you to port the hard-coded iteration order

enoriega (changes requested):
Looks very nice! Just left one simple request on `KBGenerator`

processors/src/main/scala/org/clulab/processors/bionlp/ner/KBGenerator.scala
kwalcock: I understand and expected 103, but what is "This line"?  I'm missing the antecedent.

kwalcock: Yes, thanks, that line would use the actualLabels again.

kwalcock:
These won't pass until processors has been republished.

kwalcock:
These are passing now with the updated version of processors, @enoriega.

enoriega:
Thanks @kwalcock !

PR #756, opened 2021-06-20
Remove superfluous .par

kwalcock:


kwalcock:
It looks like I didn't read the code I was replacing close enough.  Here's a small correction.

kwalcock:
Related to this, while testing fatdynet on the Mac, I found that the default .par (without any adjustments to the context and on the particular machine that does have more than 1 core) results in just one thread at a time.  It wasn't until I added code like this and set the number of threads explicitly that parallel processing was achieved.

MihaiSurdeanu:
Thanks @kwalcock !

PR #760, opened 2021-08-24
Fix access to bioresources from jar

kwalcock:


MihaiSurdeanu:
Thank you @kwalcock !!

kwalcock:
If/when the tests pass I wonder whether @bgyori can use this branch to test in his environment before I commit.  I did do an assembly use the jar as an unmanaged dependency, but still, things can be different.

bgyori:
I tried this branch and it appears to have fixed the issue, thanks a lot!

kwalcock:
FYI @enoriega, in case you have any reservations.

enoriega:
Looks great. Thanks for fixing the bug

PR #761, opened 2021-08-29
Update Docker build

bgyori:
The previous Docker build was outdated and didn't work since bintray was shut down earlier this year. This PR implements a new Dockerfile which uses Ubuntu 20.04 and updates the Java, Scala and SBT installs. It also removes the separate test Dockerfile which was only useful when processors and bioresources were in separate repositories which is not the case anymore.


MihaiSurdeanu (approved):
Thanks @bgyori !

PR #763, opened 2022-01-05
Updated Lucene Indexers

enoriega:
The file structure of PMC OA changed on Dec 21st 2021. The xml files are split into different directories. This PR updates the NxmlIndexer class to handle this new structure.
Added a new class named PubmedAbstractIndexer to create an index of pubmed abstracts akin to the full-text OA index

### DIFF
--- a/main/src/main/scala/org/clulab/reach/indexer/NxmlIndexer.scala
+++ b/main/src/main/scala/org/clulab/reach/indexer/NxmlIndexer.scala
@@ -26,8 +26,9 @@ import org.clulab.reach.utils.Preprocess
   * Last Modified: Refactor processor client to processor annotator.
   */
 class NxmlIndexer {
-  def index(mapFiles:Iterable[String], indexDir:String): Unit = {
+  def index(docs:String, mapFiles:Iterable[String], indexDir:String): Unit = {
 
+    val baseDir = new File(docs)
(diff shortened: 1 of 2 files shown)


MihaiSurdeanu:
Nice, thank you @enoriega !

For future runs of this, can you please include a README in this directory to explain what steps need to happen to index: (a) all abstracts, and (b) all OA papers.
Thanks!

@kwalcock : if this looks Ok to you, can you please merge?

main/src/main/scala/org/clulab/reach/indexer/NxmlIndexer.scala
```diff
@@ -171,7 +171,9 @@ object NxmlIndexer {
     val docsDir = props.getProperty("docs")
     val mapFile = props.getProperty("map")
 
+    println(props)
```
kwalcock, line 174: Just in case this is left over from debugging...

kwalcock (commented):
Just one question (above).

main/src/main/scala/org/clulab/reach/indexer/NxmlIndexer.scala
kwalcock: I'm assuming not.  It could be removed later if necessary.

enoriega:
@kwalcock sorry about replying late. That print statement was already there and I found it useful when running it. Thanks for reviewing!

PR #764, opened 2022-01-06
[WIP] Add FMA and Radlex ontologies

alfredomp:


alfredomp:
We'll generate a complete PR to include MFA and Radlex ontologies.

PR #777, opened 2022-11-15
This is another way to add section names without changes to processors

kwalcock:
Convert the Document to a ReachDocument wherever necessary, namely, whenever the section will need to be accessed.


kwalcock:
I like #776 better, but the serialization, etc., from there could be used here as well.

enoriega:
No longer necessary after merging #776

PR #778, opened 2023-01-05
Change to clulab dependency

kwalcock:


kwalcock:
@enoriega, some of the large resources associated with processors are being moved between domains, so this change ensures that they will still be found when the old domain disappears.  Something that isn't obvious is that the code is also now being tested at jenkins.clulab.org with an updated installation.  There may be some changes in behavior.  Please sound the alarm if you notice problems.

kwalcock:
We had only been running test and not +test to check across Scala versions.  I'm looking into it.

enoriega:
Thanks for the heads up, @kwalcock. I will update the frailty branch with this change as soon as it is merged

PR #780, opened 2023-01-16
Update processors

kwalcock:


kwalcock:
Although this passes the tests in one sbt session, it must take a huge amount of memory.  I'm giving it 56GB of memory locally and the process still gets killed.  It may be that we have to return to running it in multiple sbt passes or track down some kind of memory leak.

Issue #785, opened 2023-03-02
Triggers aren't converted to Bio- or Coref-Mentions and don't survive serialization


kwalcock:
It looks very much like any BioEventMention or CorefEventMention that is deserialized is forced to have a BioTextBoundMention or CorefTextBoundMention as a trigger: 
https://github.com/clulab/reach/blob/c7397a4b979454854f0c4a39098d2c2bb31363a3/main/src/main/scala/org/clulab/reach/mentions/serialization/json/JSONSerializer.scala#L91

https://github.com/clulab/reach/blob/c7397a4b979454854f0c4a39098d2c2bb31363a3/main/src/main/scala/org/clulab/reach/mentions/serialization/json/JSONSerializer.scala#L198

However, when the BioEventMention or CorefEventMention is created, the TextBoundMention in the trigger is not converted similarly:

https://github.com/clulab/reach/blob/c7397a4b979454854f0c4a39098d2c2bb31363a3/main/src/main/scala/org/clulab/reach/mentions/package.scala#L28

https://github.com/clulab/reach/blob/c7397a4b979454854f0c4a39098d2c2bb31363a3/main/src/main/scala/org/clulab/reach/mentions/package.scala#L65

This might cause many problems, but one is that a serialized and then deserialized Bio- or CorefEventMention will have its trigger change type from a simple TextBoundMention to a more specific one so that the round trip is essentially invalid.  A newly enabled but old test confirms this.

Those lines should probably be changed to
```scala
          m.trigger.toBioMention.asInstanceOf[BioTextBoundMention],
```
and
```scala
          m.trigger.toCorefMention.asInstanceOf[CorefTextBoundMention],
```

It would be good for others who know more about this project to confirm the intention of the original design and consider whether the change would cause problems.  Thanks.

FYI @enoriega, @MihaiSurdeanu

kwalcock:
For instance, and this was a concern in Eidos, the original TextBoundMention may be shared between two EventMentions and something may be depending on that reference equality.  That single trigger will be converted into two separate  CorefTextBoundMention copies which might be manipulated separately, for instance, by serializing them both.  They will have the same ID, leading to duplicate keys in some database, for example.

kwalcock:
Similarly, it doesn't look like the paths are converted.  I get something like a CorefEventMention but the paths it has are simple TextBoundMentions.  This can mean that the IDs don't match up.  It looks like sometimes the same mention is both an argument as a CorefTextBoundMention and in the path as a TextBoundMention.

Issue #784, opened 2023-03-01
Failing test is not actually tested


kwalcock:
The test https://github.com/clulab/reach/blob/c7397a4b979454854f0c4a39098d2c2bb31363a3/main/src/test/scala/org/clulab/reach/mentions/serialization/TestJSONSerializer.scala#L83 doesn't do anything except calculate the value `false`.  It should instead fail.  The bug that causes the false value is actually in processors and will be fixed soon.  Still, a test should test.

kwalcock:
This is coded and part of a PR.

Issue #783, opened 2023-02-22
RunReachCLI gets stuck with specific nxml files


guerrerosimonl:
I'm wondering if this is a bug. I have been running a bunch of nxml files (1240 of them) with the standard procedure described at https://github.com/clulab/reach/wiki/Running-Reach. 1224 papers are able to run successfully (re-tested in a couple of them). 16 of them stay for hours running and don't let the process finish (also tried individually). 

The 16 files are here: https://we.tl/t-1DHGfBjKc9

kwalcock:
Someone saw this and will try to replicate.

kwalcock:
I took the smallest of the files (PMC8589633.nxml), divided it up into even smaller parts, and ran them.  They all finished, eventually, like after 24 hours.   That particular file did not get stuck in an infinite loop or anything.  It's just slow.  That's partly to blame on there being around 7425 mentions found, but more to blame on some inefficient code for the serial-json format.  If you don't happen to need that format, removing it from the list is an easy solution.  If that's not possible and you have time, you might wait it out.  Be sure to allow Java lots of memory, probably more than 10GB, so that doesn't spend too much time garbage collecting.  A third alternative is to wait for some more efficient code, which is what will be discussed below.

kwalcock:
@enoriega, the main problem seems to be in `org/clulab/reach/mentions/serialization/json/package.scala` where in the process of producing IDs, things like `BioTextBoundMentionOps` are running `TextBoundMentionOps(tb).jsonAST` which calls into processors code to calculate `document.equivalenceHash`.  A hash for an entire Document is a major undertaking.  The processors code also calculates the `equivalenceHash` of the Mention itself, which in turn calls `document.equivalenceHash` again.  Then the `BioTextBoundMentionOps` replaces the id in json with yet another calculation of the Mention's ID which again calculates the document's `equivalenceHash`.  That's at least 3 times the same hash value is used per Mention, and more if that Mention is related to other mentions as triggers or arguments.  There are around 7500 Mentions in the Document, so this can take a very long time.

Processors can't in general know that the Document hasn't changed between serializations of Mentions, so the recalculation is partially justified.  Reach knows that all the mentions are being serialized at the end of processing with no further changes expected to the Document.  I believe that a cache of document equivalenceHashes can be stored there so that values can be reused.    Some code may have to be copied over from processors in order to achieve this.  (Maybe not if some related changes to processors go through.)  I'll assign this to myself in nobody objects.

[...]

PR #786, opened 2023-03-04
Speed up (and debug) serial json output

kwalcock:
This is a draft because it depends on a processors update that needs to be published and there is still a small amount of test code remaining.

The most important enhancement is that the document.equivalenceHash is now cached and is not recalculated a bajillion times.  Otherwise,

* the mentions serialization package file has been replaced by a regular file
* the mentions and related classes are not implicitly converted to MentionOps
* triggers are converted to Bio and Coref instances
* lots of duplicate code has been eliminated

This will be cleanup up and documented when processors is ready.


kwalcock:
See #783, #784, #785.

Issue #802, opened 2023-10-15
trying to integrate bionp processors from reach in odinson


Crispae:
Hi, I am trying to integrate bionlp processors from reach in odinson.
With sbt i tried to load the dependency using this

```
libraryDependencies ++= Seq("com.twosixlabs.dart" %% "odinson-api" % "3.0.16",
"org.clulab" %% "reach-main" % "1.6.2")
```
I am using odinson-api from dart , and integrating bionlp processors in it, but could not import it. While importing,
```
warn]  Note: Unresolved dependencies path:
[error] sbt.librarymanagement.ResolveException: Error downloading org.clulab:reach-processors_2.12:1.6.2
[error]   Not found
[error]   Not found
[error]   not found: /home/saurav/.ivy2/localorg.clulab/reach-processors_2.12/1.6.2/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/org/clulab/reach-processors_2.12/1.6.2/reach-processors_2.12-1.6.2.pom
[error]   not found: https://jcenter.bintray.com/org/clulab/reach-processors_2.12/1.6.2/reach-processors_2.12-1.6.2.pom
[error]   not found: http://artifactory.cs.arizona.edu:8081/artifactory/sbt-release/org/clulab/reach-processors_2.12/1.6.2/reach-processors_2.12-1.6.2.pom
[error]   not found: /home/saurav/.ivy2/local/default/org/clulab/reach-processors_2.12/1.6.2/reach-processors_2.12-1.6.2.pom
[error]         at lmcoursier.CoursierDependencyResolution.unresolvedWarningOrThrow(CoursierDependencyResolution.scala:344)
[error]         at lmcoursier.CoursierDependencyResolution.$anonfun$update$38(CoursierDependencyResolution.scala:313)
[error]         at scala.util.Either$LeftProjection.map(Either.scala:573)
[error]         at lmcoursier.CoursierDependencyResolution.update(CoursierDependencyResolution.scala:313)
[error]         at sbt.librarymanagement.DependencyResolution.update(DependencyResolution.scala:60)
[error]         at sbt.internal.LibraryManagement$.resolve$1(LibraryManagement.scala:60)
[error]         at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$12(LibraryManagement.scala:134)
[error]         at sbt.util.Tracked$.$anonfun$lastOutput$1(Tracked.scala:74)
[...]

PR #803, opened 2023-10-18
Add bioresources to build

bgyori:
This PR makes a small change to build.sbt which allows bioresources to be made available when doing `publishLocal`. This is necessary for #802.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -63,7 +63,7 @@ lazy val commonSettings = Seq(
 
 lazy val root = (project in file("."))
   .settings(commonSettings: _*)
-  .aggregate(processors, main, causalAssembly, export)
+  .aggregate(processors, main, causalAssembly, export, bioresources)
   .dependsOn(main % "test->test;compile", causalAssembly, export) // so that we can import from the console
   .settings(
     name := "reach-exe",


kwalcock:
Thanks, again, @bgyori.  I'm not sure what we should do with this.  Since it is already in a branch, perhaps we should just leave it there for safe keeping.  @enoriega may want to weigh in.

bgyori:
I think this PR is not that controversial - it just adds bioresources to be created with `publishLocal` which "doesn't hurt" so to speak, so I think it could be merged. Though I might be missing something.

enoriega:
@kwalcock I am surprised that I worked before given that the project does load resources from the `bioresources` subproject.
Regardless, if it doesn't hurt I don't object

kwalcock:
I just didn't want to presume.  The aggregation will work not just for publishLocal, but also publish, release, and test as far as I know.  The publish in particular would have consequences.

enoriega:
The subproject only consists of assets, no code whatsoever. So I am not worried about it. What I am still puzzled by if how we could read the resources and config of that subproject if it wasn't aggregated before

kwalcock:
The `dependsOn` is probably handling it.  The `processors` subproject `dependsOn bioresources`.

